PR #5, opened 2016-06-19
Switching to digishield difficulty retargeting algorithm

kaykurokawa:
Switching to digishield difficulty retargeting algorithm. Algorithm is based off of dogecoin implementation, only difference is minimum timespan factor was changed from 4 to 8. In summary the algorithm will retarget every block (2.5 min) , and difficulty will either increase by maximum of 13% or decrease by 33%

This breaks the unit test miner_tests.cpp which contains pre generated nonces to meet the bitcoin difficulty retargeting algorithm. Need to generate new nonces so the test will pass.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -127,6 +127,7 @@ BITCOIN_CORE_H = \
   miner.h \
   nameclaim.h \
   claimtrie.h \
+  lbry.h \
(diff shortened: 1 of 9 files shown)


kaykurokawa:
fixed miner_tests.cpp with new nonces

PR #11, opened 2016-07-06
zip binaries and deploy to github

jobevers:


jobevers:
The build on travis won't actually check that the release happens properly, but I did push a tag (test-release-5) that caused a successful release.

PR #12, opened 2016-07-08
Resurrect gethashespersec

kimihiro64:
The original gethashespersec rpc command was implemented in bitcoin core, but was removed in Jan 2015 shortly before they removed the internal miner. Most of the code has been copied verbatim, but there were some structural file changes, as well as api changes, that I've adjusted for. When I ran this, I got 168 KH/s on my macbook.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,8 @@ src/qt/forms/ui_*.h
 
 src/qt/test/moc*.cpp
 
+dependencies/*
+
 .deps
 .dirstamp
 .libs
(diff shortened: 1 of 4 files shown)


kaykurokawa:
untested ACK

kauffj:
@kimihiro64 could you add tests for this?

kimihiro64:
@kauffj What kind of tests are you looking for? Specifically for the rpc part, or the logic for the calculations/output? I'll look through the other tests and see if I can't figure some out.

PR #13, opened 2016-07-11
make regtest generate blocks faster so that tests run in a reasonable amount of time

jimmykiselak:
The introduction of the new retargeting function caused blocks in regtest mode to take a very long time to generate, which results in the tests taking a very very long time to run. This makes blocks in regtest generate faster.


jobevers:
Can you add a `make test` line in `reproducible_build.sh` so that they tests run as part of the build?

https://github.com/lbryio/lbrycrd/blob/master/reproducible_build.sh#L125

jobevers:
merging this so that I can add the tests to travis.

PR #14, opened 2016-07-11
Bogus claimid tests

jimmykiselak:
This adds a test which checks that someone can't just set claimId to whatever they want when creating a claim update (for example, the claimId of the current winner) and get their claim inserted immediately (or ever).


kaykurokawa:
ACK, not sure why 40735d9 is there though, that one should be merged already

PR #16, opened 2016-07-11
Make macOS build on Travis

jobevers:
Sigh.  This turned into a beast.  First, travis has a 4mb log limit which boost blasted right through.  So I piped the `make` stdout to a file.  Then, travis will timeout a build if it doesn't see anything on stdout for 10 minutes, so I had to background `make` and echo a "I'm still alive" message every minute.  Then, travis has a 50 minute timeout so I had to add caching and checks for whether the cache had been populated or not.  The staged boost installation is very large and would timeout the cache upload, so I switched over to an installed boost.

This is easily the most complicated bash script I've ever written.


reproducible_build.sh
```diff
@@ -3,27 +3,51 @@
...
+    echo "if the dependencies weren't built earlier"
+    echo
     echo "Optional arguments:"
     echo
     echo "-c: clone a fresh copy of the repo"
-    echo "-h: show help"
+    echo "-r: remove intermediate files."
+    echo "-l: build only lbrycrd"
+    echo "-d: build only the dependencies"
+    echo "-o: timeout build after 45 minutes"
```
jobevers: used on travis.  If the build doesn't stop on its own, the cache won't get populated and we'll never get this thing built.

reproducible_build.sh
```diff
@@ -56,74 +80,258 @@ if [ "$CLONE" = false ]; then
...
+    background make "${LIBEVENT_LOG}" "Waiting for libevent to finish building"
+    make install >> "${LIBEVENT_LOG}"
+}
+
+function build_dependency() {
+    PREFIX=$1
+    LOG=$2
+    BUILD=$3
+    if [ ! -d "${PREFIX}" ]; then
+	trap "cleanup \"${PREFIX}\" \"${LOG}\"" INT TERM EXIT
```
jobevers: if something fails during the build, we need to delete the build directory or else it will get added to the cache and then we'll never successfully get the dependency built

src/test/data/bitcoin-util-test.json
```diff
@@ -1,39 +1,39 @@
 [
```
jobevers: Can probably revert this file.  These tests aren't run.

jobevers:
Once this gets reviewed, I'll rebase everything down into one commit before merging.

lyoshenka:
You should run this through http://www.shellcheck.net . You don't have to fix everything because its very nitpicky, but it may have good ideas there.

Also, can you make it do the right thing by default when you run it outside of travis? I think you just have to set CLONE and TRACE to true by default.

Looking at the rest now...

reproducible_build.sh
```diff
@@ -56,74 +80,258 @@ if [ "$CLONE" = false ]; then
...
-export LIBEVENT_PREFIX="`pwd`/libevent_build"
-cd libevent
-./autogen.sh
-./configure --prefix=$LIBEVENT_PREFIX --enable-static --disable-shared --with-pic LDFLAGS="-L${OPENSSL_PREFIX}/lib/" CPPFLAGS="-I${OPENSSL_PREFIX}/include"
-make
-make install
-cd ..
+NEXT_TIME=60
+function exit_at_45() {
+    if [ -f ${HOME}/start_time ]; then
```
lyoshenka: will this pollute a user's home directory when run outside of travis?

lyoshenka: I think it may be best to have a single cleanup() function that gets run when the script exits (via trap) and does all the cleanup. Then if you have to, you can modify that function's using variables.  

Or, you could do something like http://stackoverflow.com/questions/16115144/bash-save-and-restore-trap-state-easy-way-to-manage-multiple-handlers-for-trap

jobevers: Regarding the trap, I like having it set for that dependency so that it doesn't delete the other, already completed dependencies.  I don't need to chain them because once a dependency is built I can, and do, remove the trap associated with that one.

I'll fix up the ${HOME}/start_time bit.

PR #17, opened 2016-07-12
Add osx build to travis

jobevers:
This is a large commit.  Travis has some limitations that
the OSX build runs into so a few things needed to change:

- travis has a 4mb log limit
    - log stdout/stderr to log files
    - cat relevant log file if there is a build issue
- travis will kill the job if it is silent for ten minutes
    - echo messages every minute
- travis has a 50 minute job limit
    - add caching for dependencies
    - proactively abort a build after 45 minutes to ensure the
        cache will be populated

Caching creates its own set of problems so the ability to clear the
dependency folder was added and each dependency will delete its folder
if something goes wrong during the build of that dependency. This
prevents future runs from thinking a dependency has been cached when
it actually hasn't.

The build files for OSX and linux were unified into one script.


jobevers:
This is https://github.com/lbryio/lbrycrd/pull/16, but rebased down into one commit with merge conflicts cleaned up. The intermediate commits aren't worth keeping around.

jobevers:
LGTM'd by grin in slack.

PR #22, opened 2016-07-19
Initial commit with AES algorithm

aizensou:
Hi guys.
Here is my proposal of the new AES algorithm for LBRY.
I appreciate your feedback.

Thank you.


kaykurokawa:
I don't really see a need for this.

lyoshenka:
thanks @aizensou, but we have no plans to change our algorithm at this point

PR #23, opened 2016-07-21
Limits name length of claimtrie transactions to 255 bytes

kaykurokawa:
This patch limits name length of claimtrie transactions to 255 bytes at most. 
This is a soft fork, and nodes should upgrade immediately.

### DIFF
--- a/src/nameclaim.h
+++ b/src/nameclaim.h
@@ -6,7 +6,13 @@
 
 #include <vector>
 
+// This is the max claim script size in bytes, not including the script pubkey part of the script.
+// Scripts exceeding this size are rejected in CheckTransaction in main.cpp
 #define MAX_CLAIM_SCRIPT_SIZE 8192
+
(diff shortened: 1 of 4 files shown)


src/nameclaim.h
```diff
@@ -6,15 +6,24 @@
 
 #include <vector>
 
+// This is the max claim script size, not including the script pubkey part of the script.
+// Scripts exceedeing this size are rejected in CheckTransaction in main.cpp
```
jobevers: exceedeing => exceeding

Is it obvious the units on this are bytes or should that be in the comment?

jobevers:
I don't know c++ or the lbrycrd code well enough to give a thumbs up so will defer that to somebody else.

kaykurokawa:
fixed comments, and squashed. @jimmykiselak  gave a review and Ok on slack.

PR #24, opened 2016-07-21
rpc: add claimtrie field in getblocktemplate

tpruvot:
getblocktemplate is supposed to output all required mining infos


kaykurokawa:
sounds good, ACK

PR #31, opened 2016-07-29
add windows build to travis

jobevers:
Compiles lbrycrdd using mingw to produce a windows executable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,23 @@ matrix:
           sudo: required
           dist: trusty
           language: c
+          env: TARGET=linux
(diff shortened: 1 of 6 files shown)


jobevers:
@kaykurokawa - can you double check this really quick?  The actual code changes are minor, but would like a second set of eyes just in case.

Fillerix99:
@jobevers afaik i compiled it with trevis and the linux and Windows one Works, osx throws errors.

jobevers:
the osx build errors because the boxes there are slow and the compilation takes forever.  It frequently times out.

Issue #33, opened 2016-08-07
Effective amount in getvalueforname rpc command does not return correct value 


kaykurokawa:
RPC command getvalueforname does not return correct value for "effective amount"

This is because effective amount is not serialized when serializing CClaimValue ( https://github.com/lbryio/lbrycrd/blob/master/src/claimtrie.h#L50 ). Thus reading nEffectiveAmount will not return proper value until a reorder is initiated by the claimtrie. 

Effective amount should be calculated by obtaining its supports. nEffectiveAmount should be a protected variable , with CClaimValue having CClaimTrie as a friend class.

kaykurokawa:
Fixed with https://github.com/lbryio/lbrycrd/pull/55

PR #35, opened 2016-08-08
Fix claimtrie RPC commands

kaykurokawa:
Return decimal values for amounts instead of dewey's in rpc commands getclaimsforname and getvalueforname


kaykurokawa:
Added f844880 which restructures the claimsForNameType to return more useful information. Basically moved code which was in rpc/claimtrie.cpp : getclaimsforname() that parsed the vector of claims and supports into a more useful map structure as a method in claimsForNameType so it can be used in other places. 

This also fixes https://github.com/lbryio/lbrycrd/issues/33

PR #36, opened 2016-08-08
Change rpc amounts to decimal, add getEffectiveAmountForClaim command [WIP]

kaykurokawa:
Return decimal values for amounts instead of dewey's in rpc commands getclaimsforname and getvalueforname

Add getEffectiveAmountForClaim command to use in rpc command getclaimsforname and to fix getvalueforname https://github.com/lbryio/lbrycrd/issues/33

Rework of https://github.com/lbryio/lbrycrd/pull/35 as per Jimmy's suggestions to keep view related functionality separated from claimtrie.h/.cpp

### DIFF
--- a/src/rpc/claimtrie.cpp
+++ b/src/rpc/claimtrie.cpp
@@ -183,11 +183,12 @@ UniValue getvalueforname(const UniValue& params, bool fHelp)
     std::string sValue;
     if (!getValueForClaim(claim.outPoint, sValue))
         return ret;
+
     ret.push_back(Pair("value", sValue));
     ret.push_back(Pair("txid", claim.outPoint.hash.GetHex()));
     ret.push_back(Pair("n", (int)claim.outPoint.n));
-    ret.push_back(Pair("amount", claim.nAmount));
-    ret.push_back(Pair("effective amount", claim.nEffectiveAmount)); 
+    ret.push_back(Pair("amount", ValueFromAmount(claim.nAmount)));
+    ret.push_back(Pair("effective amount", ValueFromAmount(claim.nEffectiveAmount))); 
     ret.push_back(Pair("height", claim.nHeight));
     return ret;
 }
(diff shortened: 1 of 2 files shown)


src/rpc/claimtrie.cpp
```diff
@@ -183,11 +183,14 @@ UniValue getvalueforname(const UniValue& params, bool fHelp)
...
     if (!getValueForClaim(claim.outPoint, sValue))
         return ret;
+
     ret.push_back(Pair("value", sValue));
     ret.push_back(Pair("txid", claim.outPoint.hash.GetHex()));
     ret.push_back(Pair("n", (int)claim.outPoint.n));
-    ret.push_back(Pair("amount", claim.nAmount));
-    ret.push_back(Pair("effective amount", claim.nEffectiveAmount)); 
+    ret.push_back(Pair("amount", ValueFromAmount(claim.nAmount)));
+    claimsForNameType claimsForName = pclaimTrie->getClaimsForName(name);
```
jimmykiselak: This seems like a roundabout way of calculating the effective amount. Why not have a function in CClaimTrie which calculates it on request?

kaykurokawa: You could, it would functionally be calling getClaimsForName and then performing the calculation. So was thinking its not very efficient when calculating effective amounts for multiple claims as we are doing here. 

Unless there's some better way to obtain the effective amount ?

jimmykiselak: The effective amount is only being calculated for one claim here.

As for a way to calculate the effective amounts for multiple claims, that functionality is already in place in the only places I know of that it needs to be.

jobevers:
I'd like to see a test for the new `getEffectiveAmountForClaim` function.

lyoshenka:
@jobevers this has been open since august. can we get this in for this or next cycle?

jobevers:
@kaykurokawa what is the status of this?

kaykurokawa:
I can try to get this merged in the next couple of days

kaykurokawa:
Closing as replaced by new pull request https://github.com/lbryio/lbrycrd/pull/55

PR #39, opened 2016-08-12
New claimtrie branching tests [WIP]

kaykurokawa:
Adding calimtrie branching tests. 
Also adds new claimtrie script generation functions that can be used elsewhere.

### DIFF
--- a/src/test/claimtriebranching_tests.cpp
+++ b/src/test/claimtriebranching_tests.cpp
@@ -206,6 +206,7 @@ struct ClaimTrieChainFixture{
         return tx; 
     }
 
+
     //make claim at the current block
     CMutableTransaction MakeClaim(const CTransaction &prev, std::string name, std::string value, 
                                   CAmount quantity)
@@ -233,6 +234,7 @@ struct ClaimTrieChainFixture{
         CommitTx(tx); 
         return tx;
     }
+
     //make update at the current block
     CMutableTransaction MakeUpdate(const CTransaction &prev, std::string name, std::string value,
                                uint160 claimId, CAmount quantity)
@@ -463,6 +465,7 @@ BOOST_AUTO_TEST_CASE(claimtriebranching_support)
     support spend
         spending suport on winning claim will cause it to lose
 
+        spending a support on txin[i] where i is not 0 
 */
 BOOST_AUTO_TEST_CASE(claimtriebranching_support_spend)
 {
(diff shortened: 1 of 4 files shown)


jimmykiselak:
It looks to me like this is two entirely separate PRs: one that introduces some utility functions for creating scripts for claims, and one for adding some tests. Could they be separated?

src/test/claimtriebranching_tests.cpp
```diff
@@ -0,0 +1,620 @@
...
+#include "chainparams.h"
+#include "policy/policy.h"
+#include <boost/test/unit_test.hpp>
+#include <iostream>
+#include "test/test_bitcoin.h"
+
+using namespace std;
+
+
+BOOST_FIXTURE_TEST_SUITE(claimtriebranching_tests, RegTestingSetup)
```
jimmykiselak: Is there a summary of what's tested here? What is tested here that isn't tested in claimtrie_tests.cpp?

kaykurokawa:
Yea, they are two commits so you can either git cherry-pick but I think you don't have to bother since I'm going to make a new pull request that actually uses the script generation function in place where it is needed.

So they summary of each test is in the comments above each test but I will paste them here for convenience. The goal for me was to consider each action (claim,support,spend claims, etc...) that occurs on the claimtrie and consider all the different scenario in which the action could occur, and test them individually. So for each scenario the the test gets rolled backed so we can start anew (Now that I think about it, I think I should set up separate BOOST_AUTO_TEST_CASE for each scenario, instead of separating them by action)

 The pros of this approach is that its mentally easy to keep track of what's been tested, and also when something goes wrong you can pipoint exactly what the issue is. The cons is that it will not catch sequences of events that could cause the claimtrie to fail, some of that which can be caught by claimtrie_tests.cpp . 

    claims
        no competing bids 
        there is a competing bid inserted same height
            check the greater one wins
                - quantity is same, check outpoint greater wins
        there is an existing competing bid 
            check that rules for delays are observed 
            check that a greater amount wins
            check that a smaller amount does not win    

    spent claims
        spending winning claim will make losing active claim winner    
        spending winning claim will make inactive claim winner
        spending winning claim will empty out claim trie 

    supports
        check support with wrong name does not work 
        check claim with more support wins
        check support delay 

    support spend
        spending suport on winning claim will cause it to lose

    update
        update preserves claim id
        update preserves supports 
        winning update on winning claim happens without delay
        losing update on winning claim happens without delay
        update on losing claim happens with delay , and wins

    expiration
        check claims expire and loses claim
        check claims expire and is not updateable (may be changed in future soft fork)
        check supports expire and can cause supported bid to lose claim

jimmykiselak:
claims
    no competing bids  -- https://github.com/lbryio/lbrycrd/blob/master/src/test/claimtrie_tests.cpp#L394
    there is a competing bid inserted same height -- https://github.com/lbryio/lbrycrd/blob/master/src/test/claimtrie_tests.cpp#L1522
        check the greater one wins
            - quantity is same, check outpoint greater wins
    there is an existing competing bid 
        check that rules for delays are observed -- https://github.com/lbryio/lbrycrd/blob/master/src/test/claimtrie_tests.cpp#L430
        check that a greater amount wins -- https://github.com/lbryio/lbrycrd/blob/master/src/test/claimtrie_tests.cpp#L415
        check that a smaller amount does not win -- https://github.com/lbryio/lbrycrd/blob/master/src/test/claimtrie_tests.cpp#L602


I could go on. There is no need to have two sets of tests for the same functionality. If the current tests are confusing (they are), feel free to refactor them. If the current tests are missing cases (they might be), add them. We don't need to be maintaining two sets of tests.

kaykurokawa:
agreed that there should only be one refactored file, but pushing this as low on my priority for now.

jimmykiselak:
It may be a good idea to split the file up (it is huge and takes a long time to compile), but there should only be one set of tests. Agreed on it being low priority.

jobevers:
@kaykurokawa what are your thoughts on this - still a low priority item?

kaykurokawa:
I think I would like to deprecate claimtrie_tests in favor of claimtriebraching_tests as it is easier to maintain and add new tests.

kaykurokawa:
closing as replaced by https://github.com/lbryio/lbrycrd/pull/52

PR #41, opened 2016-08-20
Helpful error messages if user confuses abandonclaim/abandonsupport

sonatagreen:
When the user tries to use abandonclaim with with a support tx, or vice versa, give an informative error message to that effect.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -592,15 +592,33 @@ void AbandonClaim(const CTxDestination &address, CAmount nAmount, CWalletTx& wtx
         throw JSONRPCError(RPC_WALLET_ERROR, "Error: The transaction was rejected! This might happen if some of the coins in your wallet were already spent, such as if you used a copy of a wallet.dat and coins were spent in the copy but not marked as spent here.");
 }
 
-UniValue abandonclaim(const UniValue& params, bool fHelp)
+UniValue abandonwhatever(const UniValue& params, bool fHelp, const isminefilter filter, bool checkOther)
 {
     if (!EnsureWalletIsAvailable(fHelp))
         return NullUniValue;
-    
+
+    string type, wrongType;
+    if (filter == ISMINE_CLAIM)
+    {
+        type = "claim";
+        wrongType = "support";
+    }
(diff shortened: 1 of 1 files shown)


kaykurokawa:
altFound looks to be an unused variable here.

sonatagreen:
Whoooops. Good catch, thanks!

sonatagreen:
Fixed.

kaykurokawa:
I'm reviewing this, and realized that there are some other issues, with the abandonclaim and supportclaim RPC command ... 

I think first of all those commands will work regardless of what you are abandoning (abandonclaim should work on supports, abandonsupport should work on claims) , the code is exactly the same, except for the isminefilter which is not used actually. 

Second , txid should not be used here as multiple claims could in theory share the same txid (and have different vouts). So this command could end up abandoning claims that you want to keep..  Ideally  we'd want to use claimId here... or at least txid + vout

sonatagreen:
I don't think abandonclaim and abandonsupport ought to be equivalent; I think it's a useful bit of syntactic salt. It might be possible to DRY out some redundant code, though.

Using claimId instead of txid is a good idea.

kaykurokawa:
Nice, I will take a look int the next couple of days, kind of strange that the travis build failed though.. doesn't seem to be related to the patch

jimmykiselak:
Unless there is some particularly good UX reason to use claimId, this should use txhash and nOut.

jimmykiselak:
I've changed my mind. Txhash and nOut should be used regardless of any ux concerns.

jimmykiselak:
Reasoning is that it's not impossible to have multiple txouts in the same tx with identical claimIDs. They could be valid supports, invalid updates, or a valid update and any number of invalid updates. In all of those cases it should be possible to uniquely specify the txout to spend.

kaykurokawa:
I agree that we should use txhash and nout. On a side note, lbryum has this same problem. 

I also don't see the purpose of having two separate function abandonclaim and abandonsupport if we are going this route. It seems like abandon is sufficient.

jobevers:
Should this be closed?

kaykurokawa:
Closing, will need to revisit when we are re writing the lbrycrd rpc calls but not high priority right now.

PR #43, opened 2016-09-08
getinfo: only report rpc errors

tpruvot:
main.cpp GetWarnings() has a special parameter for those

affect getinfo/getmininginfo errors fields

getnetworkinfo "warnings" field was left over


tpruvot:
current one reports this :

    ...
    "relayfee": 0.00001000,
    "errors": "This is a pre-release test build - use at your own risk - do not use for mining or merchant applications"
    }

kaykurokawa:
I see, I think actually fixing line 23 would be a better way to make that error go away. 

https://github.com/lbryio/lbrycrd/blob/master/src/clientversion.h

PR #45, opened 2016-09-14
fixing removesupport bug

kaykurokawa:
Fixing below exception that would occurs, when decrementing block and calling function "removeSupportFromMap" to undo a support which has been spent on a txin that is NOT txin[0] 

lbrycrdd: claimtrie.cpp:2294: bool CClaimTrieCache::decrementBlock(insertUndoType&, claimQueueRowType&, insertUndoType&, supportQueueRowType&, std::vectorstd::pair<std::basic_string<char, int> >&) const: Assertion `removeSupportFromMap(itSupportUndo->name, itSupportUndo->outPoint, support, false)' failed.

Essentially, the bug caused spent supports to not be undoable. Thus when decrementing a block where a support is spent, the support would still be spent when it should be unspent. When decrementing the block where that support was created, the removeSupportFromMap function would fail because the support its looking to remove is not there. 

With this change, clients must run -reindex once to fix the claimtrie.  

To verify the patch is working, running ./lbrycrd-cli verifychain 4 0 , should not cause an exception and return true. If not patched, running the command, will cause lbrycrdd to throw the above exception.

### DIFF
--- a/reproducible_build.sh
+++ b/reproducible_build.sh
@@ -31,6 +31,9 @@ BUILD_DEPENDENCIES=true
 BUILD_LBRYCRD=true
 TIMEOUT=false
 THREE_MB=3145728
+# this flag gets set to False if
+# the script exits due to a timeout
+OUTPUT_LOG=true
 
 while getopts :crldoth:w:d: FLAG; do
     case $FLAG in
@@ -116,6 +119,7 @@ function exit_at_45() {
 	fi
 	if [ "$TIMEOUT" = true ] && (( TIME > TIMEOUT_SECS )); then
 	    echo 'Exiting at 45 minutes to allow the cache to populate'
+	    OUTPUT_LOG=false
 	    exit 1
 	fi
     fi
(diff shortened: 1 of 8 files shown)


jobevers:
How hard would it be to add a test case that fails without this patch but passes with it

kaykurokawa:
Yea, I can work on that

kaykurokawa:
The test was added in the claimtrie branching tests https://github.com/lbryio/lbrycrd/pull/39, at https://github.com/lbryio/lbrycrd/pull/45/commits/09afe83e0c31026e59c2c1d7d9fa6e75543aa2eb  and the branch for pull request 39 was merged with this one. 

The test behaves as it should , it passes with the fix, but an exception is thrown when removing the fix commit 
https://github.com/lbryio/lbrycrd/pull/45/commits/125c4e48adbab6bfcf81d24a6cc3e87acc747d42

kaykurokawa:
Will need to seperate these as individual pull requests

kaykurokawa:
closing as replaced by https://github.com/lbryio/lbrycrd/pull/53

PR #47, opened 2016-10-13
update fixed seed nodes

lyoshenka:
lbrycrd currently uses Jack's vultr servers as the fixed seeds. This switches the mainnet seeds to LBRY servers. It also updates the testnet seed to a real testnet server


kaykurokawa:
looks good, testnet is operating with the seed node and DNS seeds for me 
ACK to merge

PR #52, opened 2016-11-01
Claimtrie branching tests with test for remove suppport bug

kaykurokawa:
This is a redo of https://github.com/lbryio/lbrycrd/pull/39, containing the new claimtrie branching test, and the test for the remove support bug. 

In the future, will need to merge claimmtrie_tests into claimtriebranching_tests to make a single test for claimtries.

### DIFF
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -61,6 +61,7 @@ BITCOIN_TESTS =\
   test/miner_tests.cpp \
   test/multisig_tests.cpp \
   test/claimtrie_tests.cpp \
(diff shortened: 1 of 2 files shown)


kaykurokawa:
The test  fails here, but this is the correct behavior as https://github.com/lbryio/lbrycrd/pull/52/commits/03b3828a659bdccca3b5be221a0ccee237469e72 should fail without  https://github.com/lbryio/lbrycrd/pull/53

jobevers:
LGTM

PR #54, opened 2016-12-21
Change regtest address related prefixes to be same as main net

kaykurokawa:
Make regtest address prefixes the same as main net. This will make coding easier to setup lbryum to work in regtest mode, which is needed in lbry-in-a-box.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -336,11 +336,12 @@ public:
             0,
             0
         };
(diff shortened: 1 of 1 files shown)


jobevers:
Not sure about this - having similar addresses increases the chances of
accidentally trying to send credits to the wrong network?  I've made the
changes needed to get lbryum configurable...

kaykurokawa:
ok i just saw that patch in lbry-in-a-box , I think that will work.

PR #56, opened 2017-01-04
Change max block decrement amount to 50

kaykurokawa:
Increase the maximum amount that a rpc call is allowed to decrement blocks from 20 to 50. This affects how far RPC command getprooforname, used by lbryum, can go back in blocks. 

Calling getproofforname and going back 50 blocks takes about 0.009 seconds (going back 20 blocks takes about 0.007) which should not be too much of a load increase.   

This should decrease cases where we get the error "Block too deep to generate proof" from lbryum when calling its getvalueforname command.

### DIFF
--- a/src/rpc/claimtrie.cpp
+++ b/src/rpc/claimtrie.cpp
@@ -4,6 +4,8 @@
 #include "univalue.h"
 #include "txmempool.h"
 
+// Maximum block decrement that is allowed from rpc calls
+const int MAX_RPC_BLOCK_DECREMENTS = 50;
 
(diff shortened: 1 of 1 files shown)


jobevers:
dumb question - what have a limit at all?

kaykurokawa:
Going back too many blocks will take too long. You don't want lbryum instances requesting the value for block index 1, causing lbrycrd to roll back thousands of blocks and holding up all other requests.   

Ideally, we should store proofs for past blocks on lbryum server , instead of getting it from lbrycrd.

PR #58, opened 2017-03-22
add claimId to return from getvalueforname

jackrobison:
Presently, to determine a claimId for a winning claim one must first `getvalueforname`, and then separately call `getclaimsfortx`.


kaykurokawa:
lgtm

PR #67, opened 2017-08-07
add getclaimbyid

jackrobison:


kaykurokawa:
Its probably worth keeping consistency with RPC command getvalueforname here, as we don't want to end up with two separate handlers in lbryum or whatever application using these two commands.

When there is no claim that exists, getvalueforname returns an empty dictionary (instead of -1). 

getvalueforname also does not format the amount value to credits.

getvalueforname also returns effective amount in credits which we can also do here.

jackrobison:
@kaykurokawa updated the return fields

kaykurokawa:
LGTM

PR #82, opened 2018-01-21
Add new transferfromaddress rpc command

### LINKED ISSUE #40: Feature Request: sendfromaddress command that keeps change in original address
**Edit: If working on this issue, please see [PR#82](https://github.com/lbryio/lbrycrd/pull/82) and either use this as a starting place or at least see the comments.**

I'd like a way to send credits from a specific address and keep the change in that address. Jack and I talked about calling it `sendfromaddress`. I'm not in love with that name (there are already several sendfrom/sendto commands) but I can't thin of a better name. This would be really helpful for maintaining public addresses that have a specific purpose.

Example:
- Address `bFromAddress123` has 5 LBC. Address `bRecipeintAddress456` is empty.
- I run `./lbrycrd-cli sendfromaddress bFromAddress123 bRecipientAddress456 2.3`
- Now `bRecipientAddress456` has 2.3LBC and `bFromAddress123` has 2.6999575LBC (or whatever `2.7 - txn fee` is).

lbrynaut:
Add new transferfromaddress rpc command that spends from a specified source address and sends any change sent back to the same address (resolves #40)

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -921,7 +921,7 @@ UniValue abandonsupport(const UniValue& params, bool fHelp)
 }
 
 
-static void SendMoney(const CTxDestination &address, CAmount nValue, bool fSubtractFeeFromAmount, CWalletTx& wtxNew)
+static void SendMoney(const CTxDestination &address, CAmount nValue, bool fSubtractFeeFromAmount, CWalletTx& wtxNew, const char* transferFromAddr)
 {
     CAmount curBalance = pwalletMain->GetBalance();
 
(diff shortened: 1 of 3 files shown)


kaykurokawa:
I don't think I'd want to alter the functionality of CreateTransaction/CommitTransaction/SendMoney for this specific custom purpose.It would be much better if we just created a custom function , that will essentially achieve what you can do through the rpc using createrawtransaction and signrawtransaction.

First reason is mainly bureaucratic in that it  now takes additional work to keep this PR and merge in changes from Bitcoin Core which has  made significant changes to the wallet code.

Second reason is that SendMoney does a lot othings that this RPC command does not need to do  i.e, reserve a key from the key pool, gather inputs. Also, the function is specifically designed to make the best effort to preserve privacy , which is the opposite of what this command does.

PR #94, opened 2018-02-10
Fix memory leak

kaykurokawa:
b2ba3f5
There was a memory leak which would cause a extra ClaimTrie node to be allocated but never deleted. 
This would happen when a new node would be created in the ClaimTrie and CClaimTrieCache.addNodeToCache() was called. 

The only reason the extra ClaimTrie node was created was to serve as a empty node template which would be copied by value over to the actual Node (I guess it looks like a DRY shortcut that was not implemented correctly). This is not necessary, and the fix just creates an empty node when necessary.

a2fa2c0
the was a missing clear() on a cache object in CClaimTrieCache. This should not cause any problem since we don't need to call clear() on map objects that are going to get destroyed anyways but nonetheless should be consistent. 

Was able to reindex with this patch, and passes tests. Uses half the memory it used to according to my tests.

### DIFF
--- a/src/claimtrie.cpp
+++ b/src/claimtrie.cpp
@@ -1220,14 +1220,24 @@ bool CClaimTrieCache::empty() const
 
 CClaimTrieNode* CClaimTrieCache::addNodeToCache(const std::string& position, CClaimTrieNode* original) const
 {
-    if (!original)
-        original = new CClaimTrieNode();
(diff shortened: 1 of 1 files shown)


kaykurokawa:
Steve has told me that he has reviewed the fix and tested it. So will be merging this.

PR #95, opened 2018-02-19
Add fees to claim trie transactions

kaykurokawa:
7f2f918 / reworked to: 1f137fc 
Add new function CalcMinClaimTrieFee to calculate the minimum fee required for claim trie transactions. Works based on MIN_FEE_PER_NAMECLAIM_CHAR , which defines the fee attached per each character of the name in the name claim. Set to 200000 base units or (0.002 LBC) for max fee 0.51 LBC (0.002*255 max length name claim). 

Added test for it in nameclaim_tests.cpp (that we should expand with other tests for nameclaim.cpp) 

8384c04 / reworked to: 12784d4
increase DEFAULT_TRANSACTION_MAXFEE from 0.1 to 1 since we can now get fees over 0.1 LBC

7f2f918 / reworked to: 5c4d8b9
implement CalcMinClaimTrieFee into AcceptToMemoryPool to allow CalcMinClaimTrieFee to take effect when rejecting transactions from the memory pool. Implement CalcMinClaimTrieFee into src/wallet/wallet.cpp so that generated transactions take into consideration the claim trie fee. 

Not sure this is the best way to do it though, ideally the function signature for AcceptToMemoryPool is not affected, but we still need a way to turn off the cacluation of minimum claim trie fee for the tests in claimtrie_tests.cpp and claimtriebranching_tests.cpp, otherwise they break. Will rethink this commit. 

EDIT2: This has been reworked : CalcMinClaimTrieFee now takes as argument minFeePerNameClaimChar . This variable is set as a global in main.cpp along with other globals. This allows it to be set freely to whatever is needed for testing purposes.

### DIFF
--- a/src/nameclaim.cpp
+++ b/src/nameclaim.cpp
@@ -1,6 +1,8 @@
+#include <boost/foreach.hpp>
 #include "nameclaim.h"
 #include "hash.h"
 #include "util.h"
+
 
 std::vector<unsigned char> uint32_t_to_vch(uint32_t n)
 {
@@ -172,4 +174,28 @@ size_t ClaimNameSize(const CScript& scriptIn)
     {
         return vvchParams[0].size();
     }
+}
+
+CAmount CalcMinClaimTrieFee(const CTransaction& tx, const CAmount &minFeePerNameClaimChar)
+{
+    if (minFeePerNameClaimChar == 0)
+    {
+        return 0;
+    }
+
+    CAmount min_fee = 0;
+    BOOST_FOREACH(const CTxOut& txout, tx.vout)
+    {
+        int op;
+        std::vector<std::vector<unsigned char> > vvchParams;
+        if (DecodeClaimScript(txout.scriptPubKey, op, vvchParams))
+        {
+            if (op == OP_CLAIM_NAME)
+            {
+                int claim_name_size = vvchParams[0].size();
+                min_fee += claim_name_size*minFeePerNameClaimChar;
+            }
+        }
+    }
+    return min_fee;
 }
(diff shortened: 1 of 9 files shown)


src/nameclaim.cpp
```diff
@@ -173,3 +175,22 @@ size_t ClaimNameSize(const CScript& scriptIn)
         return vvchParams[0].size();
     }
 }
+
+CAmount CalcMinClaimTrieFee(const CTransaction& tx, const CAmount &minFeePerNameClaimChar)
+{
+    CAmount min_fee = 0;
+    BOOST_FOREACH(const CTxOut& txout, tx.vout)
```
lbrynaut: Why not skip the loop when minFeePerNameClaimChar is 0?

src/nameclaim.h
```diff
@@ -1,33 +1,45 @@
...
 #include <vector>
 
+// This is the minimum claim fee per character in the name of an OP_CLAIM_NAME command that must
+// be attached to transactions for it to be accepted into the memory pool.
+// Rationale: current implementation of the claim trie uses more memory for longer name claims
+// due to the fact that each chracater is assigned a trie node regardless of whether it contains
+// any claims or not. In the future, we can switch to a radix tree implementation where
+// empty nodes do not take up any memory and the minimum fee can be priced on a per claim
+// basis.
+#define MIN_FEE_PER_NAMECLAIM_CHAR 200000
```
lbrynaut: The PR looks good, and I haven't had as much time to consider the fee structure, but this strikes me as a bit high for a minimum at this point ... although it'll effectively act as a deterrent (perhaps the intended effect).

src/nameclaim.cpp
kaykurokawa: good idea

tzarebczan:
@kaykurokawa what's the next step here?

kaykurokawa:
Made small change that lbrynaut suggested, rebased and merging.

PR #96, opened 2018-02-19
CClaimTrieNode Leak Fixes and Refactor

technovelist:
I have rewritten the CClaimTrieNode class in the manager/worker idiom. This means that the CClaimTrieNode objects in the code are actually manager objects that forward all functions to an embedded CClaimTrieNode_Impl object. The purpose of the manager class is to manage the life cycle of the actual worker objects so that memory leaks are prevented.

Note that this means that "new" and "delete" are no longer used in the application layer (claimtrie.cpp), and that referring to a node uses . notation rather than ->.

You can create a new node via a statement like "CClaimTrieNode x;", or clone an existing one by a statement like "CClaimTrieNode x = y.clone();". You can also copy CClaimTrieNodes and pass them as arguments or return values without having to worry about memory allocation errors.

You still do have to erase CClaimTrieNodes from maps if they are no longer in use.


kaykurokawa:
There's some good stuff in here  but let's beak it up a bit into seperate pull requests first since there are some unrelated changes. 

First , I see you added some build related scritps in there, and made some changes to the Makefile. Can you describe what you changed/added and the reasoning for your change ? Do we need to incorporate these changes, or were these just useful for your own purpose.  

Breaking out the various defined types into seperate files src/claimtriemaptypes.h and  src/claimtriemaptypes.h are  needed changes , let's break this up into seperate PR. I think these changes break the tests though, so we'll need to adjust them ( test are located in src/tests and can be run from src/tests/test_lbcyrcrd ). I will do this, but for future reference please check out some references on how to "rebase" and split up your commits in git: 

https://stackoverflow.com/questions/6217156/break-a-previous-commit-into-multiple-commits 
https://johnkary.net/blog/git-add-p-the-most-powerful-git-feature-youre-not-using-yet/

technovelist:
Sounds good but I'm going to be out most of the day. I'll be back this evening. 

Steve Heller

technovelist:
On Mon, 19 Feb 2018 18:32:17 +0000 (UTC), Umpei Kay Kurokawa <notifications@github.com> wrote:

>There's some good stuff in here  but let's beak it up a bit into seperate pull requests first since there are some unrelated changes. 

So much for the theory that no one was going to look at what I posted, so I didn't have to worry about what condition it was in. :-)

>First , I see you added some build related scritps in there, and made some changes to the Makefile. Can you describe what you changed/added and the reasoning for your change ? Do we need to incorporate these changes, or were these just useful for your own purpose.  

I think we need a build script that keeps debug information, because that is very helpful when debugging.
As for the makefile, I have added C++11 compiler options but commented them out for the moment. Those options will be needed to use my virtual memory
library, which I expect will be the case shortly.

>Breaking out the various defined types into seperate files src/claimtriemaptypes.h and  src/claimtriemaptypes.h are  needed changes , let's break this up into seperate PR. I think these changes break the tests though, so we'll need to adjust them ( test are located in src/tests and can be run from src/tests/test_lbcyrcrd ). 

If "the tests" are the 205 tests referred to in the build script, they all pass, or did when I last built. I wouldn't have considered submitting it
before that.

>I will do this, but for future reference please check out some references on how to "rebase" and split up your commits in git: 

>https://stackoverflow.com/questions/6217156/break-a-previous-commit-into-multiple-commits 
>https://johnkary.net/blog/git-add-p-the-most-powerful-git-feature-youre-not-using-yet/

Sure.
------------
Steve Heller

lbrynaut:
Closing temporarily to re-create this PR from Steve's fork

PR #99, opened 2018-02-23
Just the correct files

technovelist:


lbrynaut:
A re-opened version of #96 from a proper fork instead of lbryio/lbrycrd branch

PR #100, opened 2018-02-26
Claimtrie Node Pimpl Refactor / Move custom datatype to seperate files

kaykurokawa:
Cleaned up version of https://github.com/lbryio/lbrycrd/pull/99. This just removes some unnecessary scripts and some lines of code that were commented out.

This refactors CClaimTrieNode so that allocation/deletion of CClaimTrieNode is handled by a single class at fixed locations, and can be easily modified in the future to use custom allocators and implement disk storage/retreival of CClaimTrieNodes

The numerous datatypes defined in the claimtrie.h has also been moved to claimtrietypedefs.h and claimtriemaptypes.h . 

Comments:
I think  it would be ok if we merged claimtrietypedefs.h and claimtriemaptypes.h into one file. 

I'm getting segfault errors when I run the tests ( src/test/test_lbrycrd ) from claimtrie_tests, there seems to be some problem still.

unknown location(0): fatal error in "rpc_addmultisig": signal: SIGABRT (application abort requested)
test/claimtrie_tests.cpp(216): last checkpoint
test_lbrycrd: key.cpp:305: void ECC_Start(): Assertion `secp256k1_context_sign == NULL' failed.
unknown location(0): fatal error in "rpc_wallet": signal: SIGABRT (application abort requested)
test/claimtrie_tests.cpp(216): last checkpoint

*** 122 failures detected in test suite "Bitcoin Test Suite"
Database handles still open at environment close
Open database handle: unnamed/wallet.dat
test_lbrycrd: /usr/include/boost/thread/pthread/condition_variable_fwd.hpp:81: boost::condition_variable::~condition_variable(): Assertion `!ret' failed.
Aborted (core dumped)


technovelist:
test/claimtrie_tests.cpp(224): error: in "claimtrie_tests/claimtrie_merkle_hash": check hash2check == hash2 has failed

This is because the call "ntState.removeClaimFromTrie(std::string("testtesttesttest"), tx5OutPoint, unused);" is not updating the hash code correctly.

tzarebczan:
@kaykurokawa what's the next step here?

Issue #101, opened 2018-02-28
Need to make claims (update/ and name claim) in hex


kaykurokawa:
We need a way for users to input hex claims value in lbryrd-cli / lbrycrd RPC (command claimname and updateclaim) so that protobuf claims can be made through lbrycrd 

It's probably sufficient to accept hex as default.

PR #103, opened 2018-03-05
RPC commands claimname and updateclaim now take hex encoded "value" argument

kaykurokawa:
Specify value field in RPC command claimname and updateclaim with hex encoded decimal strings. Also removed some whitespaces at end of line in file rpcwallet.cpp while I was doing this.

https://github.com/lbryio/lbrycrd/issues/101


eukreign:
Seems to work, already using this in `lbrytest`.

Issue #10, opened 2016-07-05
Build error on MacOS


ELM4Ever:
MacOS 10.11.5

./autogen.sh and ./configure went well.  With "make" - I am getting the following error:

Making all in src
  CXX      crypto/libbitcoinconsensus_la-hmac_sha256.lo
  CXX      crypto/libbitcoinconsensus_la-hmac_sha512.lo
  CXX      crypto/libbitcoinconsensus_la-ripemd160.lo
  CXX      crypto/libbitcoinconsensus_la-sha1.lo
  CXX      crypto/libbitcoinconsensus_la-sha256.lo
  CXX      crypto/libbitcoinconsensus_la-sha512.lo
  CXX      libbitcoinconsensus_la-arith_uint256.lo
  CXX      consensus/libbitcoinconsensus_la-merkle.lo
  CXX      libbitcoinconsensus_la-hash.lo
  CXX      primitives/libbitcoinconsensus_la-block.lo
  CXX      primitives/libbitcoinconsensus_la-transaction.lo
  CXX      libbitcoinconsensus_la-pubkey.lo
  CXX      script/libbitcoinconsensus_la-bitcoinconsensus.lo
  CXX      script/libbitcoinconsensus_la-interpreter.lo
  CXX      script/libbitcoinconsensus_la-script.lo
  CXX      script/libbitcoinconsensus_la-script_error.lo
  CXX      libbitcoinconsensus_la-uint256.lo
  CXX      libbitcoinconsensus_la-utilstrencodings.lo
gcc -I. -g -O2 -Wall -Wextra -Wno-unused-function -c src/gen_context.c -o gen_context.o
gcc gen_context.o -o gen_context
./gen_context
  CC       src/libsecp256k1_la-secp256k1.lo
  CCLD     libsecp256k1.laCXXLD    libbitcoinconsensus.la
clang: warning: argument unused during compilation: '-pthread'
clang: warning: argument unused during compilation: '-pthread'
Undefined symbols for architecture x86_64:
  "memory_cleanse(void*, unsigned long)", referenced from:
      CBlockHeader::GetPoWHash() const in libbitcoinconsensus_la-block.o
[...]

PR #105, opened 2018-03-13
WIP: Unit test for recursivePruneName and for future CClaimTrieCache unit tests

kaykurokawa:
This adds a unit test for function CClaimTrieCache::recursivePruneName()

The new test file claimtriecache_tests.cpp can also contain additional unit tests for CClaimTrieCache in the future which is very lacking as of now.

I had to change use of "private" in CClaimTrieCache to "protected" in order to unit test non-public functions. Not sure if this is the best way but I think its ok. Other classes like CClaimTrie could switch to using "protected" in the future as well for better unit tests.

### DIFF
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -61,6 +61,7 @@ BITCOIN_TESTS =\
   test/miner_tests.cpp \
   test/multisig_tests.cpp \
   test/claimtrie_tests.cpp \
+  test/claimtriecache_tests.cpp \
   test/claimtriebranching_tests.cpp \
   test/netbase_tests.cpp \
   test/pmt_tests.cpp \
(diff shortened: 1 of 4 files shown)


kaykurokawa:
I added simple tests here for computing merkle hashes on the trie which could be expanded upon. 
In this case it tests the hash for when the trie is empty, and when it just has a single empty root node.

technovelist:
Please generate at least test #10 of the tests I asked for. Here's the list again:

I need some tests for the claimtrie rewrite project, to get a baseline on the existing code's behavior.

Here is a preliminary list. In each case, the test should complete the block, persist it, and display the hash for each node in the block so that I can try to reproduce the hashing.

If any of these tests are impossible to write, please let me know why.

1. A block with no transactions.

2. A block with 1 transaction, 0 claims.
3. A block with 1 transaction, 1 claim.
4. A block with 1 transaction, 2 claims. The first claim should be the winner.
5. A block with 1 transaction, 2 claims. The second claim should be the winner.

6. A block with 2 transactions that do not share any part of the key. 0 claims in either transaction.
7. A block with 2 transactions that do not share any part of the key. 1 claim in the first transaction, 0 in the second.
8. A block with 2 transactions that do not share any part of the key. 0 claim in the first transaction, 1 in the second.

9. A block with 2 transactions that share the first byte of the key. 0 claims in either transaction.
10. A block with 2 transactions that share the first byte of the key. 1 claim in the first transaction, 0 in the second.
11. A block with 2 transactions that share the first byte of the key. 0 claim in the first transaction, 1 in the second.

PR #112, opened 2018-04-01
Implement a hard fork for extended/infinite claim expiration times

lbrynaut:


kaykurokawa:
Nice.

I think we first need to get a version of this that is not mainnet altering so that downstream LBRY devs can finally test expiration better using regtest (they haven't been able to because of the long expiration time and generating that many blocks take too long).

I made some alterations on this branch: https://github.com/lbryio/lbrycrd/commits/claim-expiration

This branch changes mainnet chainparams so that there should be no forking change in expiration
https://github.com/lbryio/lbrycrd/commit/3539af2a2bc419bc0ca66121efd3bb3f7776a554

I made it so that it claim expiration is not disabled by default, and utilized a #define statement, I made the if statement a bit easier to understand by using nCurrentHeight on CClaimTrie
https://github.com/lbryio/lbrycrd/commit/c3646afb63eb378d31eabc87d738d78552ea9c7a

I was succesfully able to resync to the main-net with these changes. Let me know what you think.

lbrynaut:
@kaykurokawa Thanks for looking at this!

Explicitly disabling on mainnet is a fine idea so long as we're not looking to hard fork.  The testnet/regtest params were already independent, so they could be tested at any reasonable heights.  I agree picking an arbitrary mainnet fork height needs discussion, so for now, disabling is a fair option to move ahead with any testing required.

kaykurokawa:
Closing this and continuing here: https://github.com/lbryio/lbrycrd/pull/115

PR #115, opened 2018-04-04
Implement capability for a hard fork for extended/infinite claim expiration times 

kaykurokawa:
Based on https://github.com/lbryio/lbrycrd/pull/112, unlike this one, this PR should not hard fork mainnet or testnet. It does implement the hard forking in regtest so it can be tested downstream. Regtest sets original expiration time to 500, forks at height 800 and sets it to new expiration time 600. 

Below are additions to  https://github.com/lbryio/lbrycrd/pull/112:

Made #define statement to change whether we are removing claim expiration or not, currently set to not remove it. 

Adjusted chainparams so mainnet and testnet does not hard fork, made regtest fork to an extended expiration time for testing purposes. 

Changed the initialization of nExpirationTime in CClaimTrie to use chainparams 

I adjusted claimtriebranching_no_expire  to account for the fact that CClaimTrie should be using RegTesting expiration related parameters (so no need to set nExpirationTime inside the test). Also added a test that checks to see if claims that expire after the fork height still expires at the original height as expected.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -132,8 +132,8 @@ public:
         consensus.nPowTargetTimespan = 150; //retarget every block
         consensus.nPowTargetSpacing = 150;
         consensus.nOriginalClaimExpirationTime = 262974;
-        consensus.nExtendedClaimExpirationTime = 2102400;
-        consensus.nExtendedClaimExpirationForkHeight = 400000; // FIXME: pick a real height
+        consensus.nExtendedClaimExpirationTime = 262974;
+        consensus.nExtendedClaimExpirationForkHeight = -1; // FIXME: pick a real height
         consensus.fPowAllowMinDifficultyBlocks = false;
         consensus.fPowNoRetargeting = false;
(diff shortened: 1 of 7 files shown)


lbrynaut:
Why an entirely new PR?

Also, I get the impression no one on this team really squashes commits, but I would urge you to squash those into 1 on top of mine.

kaykurokawa:
Sorry, I guess I didn't have to open a new PR for this. In the other lbry repos, there's various build related things that don't run if the PR is not on a lbryio branch (if they are from someone else) due to security reasons but lbrycrd doesn't have this. Next time i'll just point you to the new commits.  

I generally keep the commits unsquashed to make it easier for the reviewer. But yes, I have a bad habit of leaving them unsquashed and should squash them once its finished with review.  Thanks.

lbrynaut:
Ok, just curious.  I would argue that unsquashed commits make reviewing potentially much more difficult, as it's very possible to review a large commit that is invalidated by a later commit.  Reviewing the final code is all that matters.

kaykurokawa:
merged as part of https://github.com/lbryio/lbrycrd/pull/137

PR #118, opened 2018-04-11
use specific version of libevent in reproducible builds

kaykurokawa:
In reproducible_build.sh, to build libevent, it would just download from the master branch of libveven'ts github repo without checking out any specific version. 

Download 2.0.22 release instead as specified here: https://github.com/lbryio/lbrycrd/blob/master/depends/packages/libevent.mk


lyoshenka:
@kaykurokawa whats next for this PR? does it need to be reviewed?

kaykurokawa:
No will just merge this , the build problem on travis with this PR is unrelated it seems.

PR #123, opened 2018-04-21
Test and fix CClaimTrie::getEffectiveAmountForClaim

shyba:
Fixed a missing block after the if statement and added some more tests, they fail without the fix.


kaykurokawa:
yeesh, nice find.
Looks good to merge.

PR #135, opened 2018-05-07
Small fix for best_claim_effective_amount_equals 

kaykurokawa:
Noticed that in test fixture best_claim_effective_amount_equals, we called getEffectiveAmountForClaim() but never used its result. This is likely why the error in this PR https://github.com/lbryio/lbrycrd/pull/123 was never caught. 

There is not really any functional difference between the original code and this change. 

Noting that a small minor difference is that the effective amount returned in getInfoForName claim value is calculated on demand and may not necessarily be the correct value. For example upon lbrycrdd restart, the effective amount returned by getInfoForName will be garbage because it hasn't been calculated yet.  This has no impact on these unit tests because we do not test restarts.


shyba:
@kaykurokawa LGTM. Nice catch!

PR #110, opened 2018-03-26
Add a claim index for fast reference by id

lbrynaut:

PR #137, opened 2018-05-10
Add capability to extend expiration time of pre fork claims

kaykurokawa:
**Details of hard fork for all:**

The fork of the mainnet will be on block 400155 (will be around noon EST 7/9/18) 

The fork of the testnet will be on block 280831 (will be around noon EST 5/21/18)

Claims made on and after the target hardfork blocks will expire after 2102400 blocks (approximately 10 years assuming block time 2.5 minutes).

Claims that were made before the hardfork and have not expired will have their expiration extended by the difference between the new expiration time and the previous expiration time (456 days). 

Claims that already expired will remain expired. 

**Details for reviewers**

This is further work off of : https://github.com/lbryio/lbrycrd/pull/110 in order to add the extension of claim expiration to already existing claims ( in #110 claims made before the hard fork would still expire at the old expiration time)

While working on this, I made two adjustments to #110. A) the first was that the new expiration time took effect 1 block after the target hard fork block, and not on the target hard fork block. It would be more intuitive if the new expiration time took effect on the target hard fork block. B) I fixed the logic of calling setExpirationTime() to only be called when we are at the target hard fork block, instead of on every block. This is more intuitive and in line with the soft-fork logic that is already present for Bitcoin. This means we need to also called setExpirationTime when we load the claimtrie from disk so I did that.  The combination of A) and B) fixes a bug in #110 where if you loaded the claimtrie from disk (restarted lbrycrdd) than you expiration time for the first block after restart would be the old expiration time instead of the new one even after the hard fork. 

The above work is on https://github.com/lbryio/lbrycrd/pull/137/commits/2df3bb104b29f5f1cd0615efbe33ab46bc1aca53

While working on extending the expiration of existing claims, I realized that #110 had the problem where it would not properly be able to remove expirations from the expiration queue properly.  This is because the function removeFromExpirationQueue() : https://github.com/lbryio/lbrycrd/blob/master/src/claimtrie.cpp#L1675 was not adjusted to consider the hard fork change in expiration time hence this line: https://github.com/lbryio/lbrycrd/blob/master/src/claimtrie.cpp#L1677 could calculate the wrong expiration time for a claim that was created before the hard fork an expires after the hard fork. 

The way we extend the claim is fairly straight forward. When we encounter the hard fork block, we set the new expiration time, and we look at all the entries in the expiration queue either in the dirty queue (where entries that are not saved to disk resides) or the disk database. The entries are adjusted to have their expiration time extended by the difference between the new expiration time and the old expiration time. Effectively, this makes it look like all the claims share the same new expiration time so the issue I described is solved. 

The above work on extending the expiraiton of existing claims is on 
b3aa62c 

For the unit tests, I added a series of tests to see if supports behaved properly, and also added tests to see if the claimtrie behaved properly on the hard fork if I simulated a restart (write claimtrie to disk, clear it in memory, and than read it from disk). https://github.com/lbryio/lbrycrd/pull/137/commits/6621f50b985d49a5b5c4c09b32d871c58fa54121

The other commits should be straight forward.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -131,6 +131,9 @@ public:
         consensus.powLimit = uint256S("0000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.nPowTargetTimespan = 150; //retarget every block
         consensus.nPowTargetSpacing = 150;
+        consensus.nOriginalClaimExpirationTime = 262974;
+        consensus.nExtendedClaimExpirationTime = 2102400;
+        consensus.nExtendedClaimExpirationForkHeight = 400155;
         consensus.fPowAllowMinDifficultyBlocks = false;
         consensus.fPowNoRetargeting = false;
         consensus.nRuleChangeActivationThreshold = 1916; // 95% of 2016
@@ -215,6 +218,9 @@ public:
         consensus.powLimit = uint256S("0000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.nPowTargetTimespan = 150;
         consensus.nPowTargetSpacing = 150;
+        consensus.nOriginalClaimExpirationTime = 262974;
+        consensus.nExtendedClaimExpirationTime = 2102400;
+        consensus.nExtendedClaimExpirationForkHeight = 278160;
         consensus.fPowAllowMinDifficultyBlocks = true;
         consensus.fPowNoRetargeting = false;
         consensus.nRuleChangeActivationThreshold = 1512; // 75% for testchains
@@ -292,6 +298,9 @@ public:
         consensus.powLimit = uint256S("7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.nPowTargetTimespan = 1;//14 * 24 * 60 * 60; // two weeks
         consensus.nPowTargetSpacing = 1;
+        consensus.nOriginalClaimExpirationTime = 500;
+        consensus.nExtendedClaimExpirationTime = 600;
+        consensus.nExtendedClaimExpirationForkHeight = 800;
         consensus.fPowAllowMinDifficultyBlocks = false;
         consensus.fPowNoRetargeting = false;
         consensus.nRuleChangeActivationThreshold = 108; // 75% for testchains
--- a/src/claimtrie.cpp
+++ b/src/claimtrie.cpp
@@ -218,6 +218,7 @@ bool CClaimTrie::supportQueueEmpty() const
 void CClaimTrie::setExpirationTime(int t)
 {
     nExpirationTime = t;
+    LogPrintf("%s: Expiration time is now %d\n", __func__, nExpirationTime);
 }
 
 void CClaimTrie::clear()
@@ -1119,6 +1120,8 @@ bool CClaimTrie::ReadFromDisk(bool check)
         LogPrintf("%s: Couldn't read the best block's hash\n", __func__);
     if (!db.Read(CURRENT_HEIGHT, nCurrentHeight))
         LogPrintf("%s: Couldn't read the current height\n", __func__);
+    setExpirationTime(Params().GetConsensus().GetExpirationTime(nCurrentHeight-1));
+
     boost::scoped_ptr<CDBIterator> pcursor(const_cast<CDBWrapper*>(&db)->NewIterator());
 
     pcursor->SeekToFirst();
@@ -1699,8 +1702,17 @@ bool CClaimTrieCache::removeClaim(const std::string& name, const COutPoint& outP
 
 void CClaimTrieCache::addToExpirationQueue(int nExpirationHeight, nameOutPointType& entry) const
 {
-    expirationQueueType::iterator itQueueRow = getExpirationQueueCacheRow(nExpirationHeight, true);
-    itQueueRow->second.push_back(entry);
+    // NOTE: To disable expiration completely after fork, set this define to enable check
+    // #define CLAIM_EXPIRATION_DISABLED_AFTER_FORK
+    #ifdef CLAIM_EXPIRATION_DISABLED_AFTER_FORK
+    const Consensus::Params& consensusParams = Params().GetConsensus();
+    if ((nExpirationHeight < consensusParams.nExtendedClaimExpirationForkHeight) ||
+        (base->nCurrentHeight < consensusParams.nExtendedClaimExpirationForkHeight))
+    #endif
+    {
+        expirationQueueType::iterator itQueueRow = getExpirationQueueCacheRow(nExpirationHeight, true);
+        itQueueRow->second.push_back(entry);
+    }
 }
 
 void CClaimTrieCache::removeFromExpirationQueue(const std::string& name, const COutPoint& outPoint, int nHeight) const
@@ -1715,10 +1727,11 @@ void CClaimTrieCache::removeFromExpirationQueue(const std::string& name, const C
             if (name == itQueue->name && outPoint == itQueue->outPoint)
                 break;
         }
-    }
-    if (itQueue != itQueueRow->second.end())
-    {
-        itQueueRow->second.erase(itQueue);
+
+        if (itQueue != itQueueRow->second.end())
+        {
+            itQueueRow->second.erase(itQueue);
+        }
     }
 }
 
@@ -2140,6 +2153,7 @@ bool CClaimTrieCache::incrementBlock(insertUndoType& insertUndo, claimQueueRowTy
             CClaimValue claim;
             assert(removeClaimFromTrie(itEntry->name, itEntry->outPoint, claim, true));
             expireUndo.push_back(std::make_pair(itEntry->name, claim));
+            LogPrintf("Expiring claim %s: %s, nHeight: %d, nValidAtHeight: %d\n", claim.claimId.GetHex(), itEntry->name, claim.nHeight, claim.nValidAtHeight);
         }
         itExpirationRow->second.clear();
     }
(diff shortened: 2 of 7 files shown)


src/chainparams.cpp
```diff
@@ -131,6 +131,9 @@ class CMainParams : public CChainParams {
         consensus.powLimit = uint256S("0000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.nPowTargetTimespan = 150; //retarget every block
         consensus.nPowTargetSpacing = 150;
+        consensus.nOriginalClaimExpirationTime = 262974;
+        consensus.nExtendedClaimExpirationTime = 2102400;
+        consensus.nExtendedClaimExpirationForkHeight = 400155; // FIXME: pick a real height
```
lbrynaut: Remove FIXME comment

src/chainparams.cpp
```diff
@@ -292,6 +298,9 @@ class CRegTestParams : public CChainParams {
         consensus.powLimit = uint256S("7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.nPowTargetTimespan = 1;//14 * 24 * 60 * 60; // two weeks
         consensus.nPowTargetSpacing = 1;
+        consensus.nOriginalClaimExpirationTime = 500;
+        consensus.nExtendedClaimExpirationTime = 600;
+        consensus.nExtendedClaimExpirationForkHeight = 800; // FIXME: pick a real height
```
lbrynaut: Remove FIXME comment

src/claimtrie.cpp
```diff
@@ -2563,3 +2571,106 @@ CClaimTrieProof CClaimTrieCache::getProofForName(const std::string& name) const
...
+    for (expirationQueueType::const_iterator i = base->dirtySupportExpirationQueueRows.begin(); i != base->dirtySupportExpirationQueueRows.end(); ++i)
+    {
+        int height = i->first;
+        dirtySupportHeights.insert(height);
+        expirationQueueRowType row = i->second;
+        removeAndAddSupportToExpirationQueue(row, height, increment);
+    }
+
+
+    //look through db for expiration queues, if we haven't already found it in dirty expiraiton queue
```
lbrynaut: expiration is spelled incorrectly in several places, maybe do a quick search for them?

src/main.cpp
```diff
@@ -2171,12 +2171,22 @@ bool DisconnectBlock(const CBlock& block, CValidationState& state, const CBlockI
                     if (op == OP_CLAIM_NAME)
                     {
                         assert(vvchParams.size() == 2);
+                        std::string name(vvchParams[0].begin(), vvchParams[0].end());
+                        std::string value(vvchParams[1].begin(), vvchParams[1].end());
                         claimId = ClaimIdHash(hash, i);
+                        LogPrintf("--- %s[%lu]: OP_CLAIM_NAME \"%s\" = \"%s\" with claimId %s and tx prevout %s at index %d\n",
+                                  __func__, view.AccessCoins(hash)->nHeight, name, SanitizeString(value),
```
lbrynaut: I realize I initially used the view height, but we could replace with pindex->nHeight (in all places it's used in debug)

src/main.cpp
```diff
@@ -2220,6 +2232,13 @@ bool DisconnectBlock(const CBlock& block, CValidationState& state, const CBlockI
     assert(trieCache.finalizeDecrement());
     trieCache.setBestBlock(pindex->pprev->GetBlockHash());
     assert(trieCache.getMerkleHash() == pindex->pprev->hashClaimTrie);
+    if (pindex->nHeight == Params().GetConsensus().nExtendedClaimExpirationForkHeight)
+    {
+        LogPrintf("Decremented past the fork v13 fork height");
```
lbrynaut: What is v13 fork?  Why not call it the claim expiration fork or something more intuitive?

src/test/claimtriebranching_tests.cpp
```diff
@@ -18,12 +18,12 @@
 #include <iostream>
 #include "test/test_bitcoin.h"
 
+
```
lbrynaut: ?

src/test/claimtriebranching_tests.cpp
```diff
@@ -164,14 +164,26 @@ struct ClaimTrieChainFixture{
...
         fRequireStandard = false;
         BOOST_CHECK(pclaimTrie->nCurrentHeight == chainActive.Height() + 1);
         LOCK(cs_main);
         num_txs_for_next_block = 0;
         num_txs = 0; 
         coinbase_txs_used = 0;
+
+
+
         // generate coinbases to spend
```
lbrynaut: Seems 1 line is enough?

src/test/claimtriebranching_tests.cpp
```diff
@@ -300,6 +312,17 @@ struct ClaimTrieChainFixture{
         num_txs_for_next_block = 0; 
     }
 
+
+    void WriteClearReadClaimTrie()
+    {
+        // this will simulate restart of lbrycrdd by writing the claimtrie to disk,
+        // clearing the-in memory claimtrie, and then reading it the saved claimtrie
```
lbrynaut: s/reading it/reading/

src/test/claimtriebranching_tests.cpp
```diff
@@ -300,6 +312,17 @@ struct ClaimTrieChainFixture{
...
+        // from disk
+        pclaimTrie->WriteToDisk();
+        pclaimTrie->clear();
+        pclaimTrie->ReadFromDisk(true);
+    }
```
lbrynaut: This also doesn't do exactly what's in the comment since the Cache isn't modified/updated/rebuilt.

kaykurokawa:
It was discovered that the testnet blocks haven't been progressing (mining was turned off) so I'll need to set a new fork height for the testnet.

src/test/claimtriebranching_tests.cpp
kaykurokawa: What do you mean here exactly ? You may have misssed where WriteToDisk clears out the cache entries https://github.com/lbryio/lbrycrd/blob/master/src/claimtrie.cpp#L1047 here

kaykurokawa:
made fixes as reviewed by lbrynaut

src/test/claimtriebranching_tests.cpp
lbrynaut: I was referring to the state of the TrieCache.

src/main.cpp
```diff
@@ -2171,12 +2171,22 @@ bool DisconnectBlock(const CBlock& block, CValidationState& state, const CBlockI
                     if (op == OP_CLAIM_NAME)
                     {
                         assert(vvchParams.size() == 2);
+                        std::string name(vvchParams[0].begin(), vvchParams[0].end());
+                        std::string value(vvchParams[1].begin(), vvchParams[1].end());
                         claimId = ClaimIdHash(hash, i);
+                        LogPrintf("--- %s[%lu]: OP_CLAIM_NAME \"%s\" = \"%s\" with claimId %s and tx prevout %s at index %d\n",
```
shyba: maybe log this only if in debug mode? otherwise `getnameproof` can become verbose as it currently disconnects and reconnects blocks for generating previous block proofs.

src/test/claimtriebranching_tests.cpp
```diff
@@ -796,4 +816,204 @@ BOOST_AUTO_TEST_CASE(claimtriebranching_get_effective_amount_for_claim)
...
+    CMutableTransaction tx2 = fixture.MakeClaim(fixture.GetCoinbase(),"test2","one",1);
+    CMutableTransaction s2 = fixture.MakeSupport(fixture.GetCoinbase(),tx2,"test2",1);
+    fixture.IncrementBlocks(1);
+    fixture.WriteClearReadClaimTrie();
+    CMutableTransaction u2 = fixture.MakeUpdate(tx2,"test2","two",ClaimIdHash(tx2.GetHash(),0),1);
+    // increment to fork
+    fixture.IncrementBlocks(fixture.expirationForkHeight - chainActive.Height());
+    BOOST_CHECK(is_best_claim("test2", u2));
+    BOOST_CHECK(best_claim_effective_amount_equals("test2",2));
+    // increment to original expiraiton, should not be expired
```
shyba: s/expiraiton/expiration/

src/test/claimtriebranching_tests.cpp
```diff
@@ -796,4 +816,204 @@ BOOST_AUTO_TEST_CASE(claimtriebranching_get_effective_amount_for_claim)
...
+    CMutableTransaction tx2 = fixture.MakeClaim(fixture.GetCoinbase(),"test2","one",3);
+    fixture.IncrementBlocks(1);
+    BOOST_CHECK_EQUAL(chainActive.Height(), fixture.expirationForkHeight -1);
+
+    // Disable future expirations and fast-forward past the fork height
+    fixture.IncrementBlocks(1);
+    BOOST_CHECK_EQUAL(chainActive.Height(), fixture.expirationForkHeight);
+    BOOST_CHECK_EQUAL(pclaimTrie->nExpirationTime, fixture.extendedExpiration);
+    // make sure decrementing to before the fork height will apppropriately set back the
+    // expiration time to the original expiraiton time
```
shyba: s/expiraiton/expiration/

src/test/claimtriebranching_tests.cpp
shyba: oh, I see @lbrynaut  already found that, sorry for the flood.

src/main.cpp
kaykurokawa: is there a debugging log?

src/test/claimtriebranching_tests.cpp
kaykurokawa: fixed

kaykurokawa:
Made fix as suggested by syba, 
Adjusted fork height on testnet to 278080 (in about an hour from now) , and I will run this on testnet.
Maybe for debug logging comment by shyba, we can do another PR to perhaps improve/clean up different log messages?

shyba:
@kaykurokawa yes, I agree. Created an [issue for logging here](https://github.com/lbryio/lbrycrd/issues/144).

kaykurokawa:
Rebased to two commits, rebased to master

Issue #136, opened 2018-05-09
Disk related functions should be seperated out from ClaimTrie class into its own class


kaykurokawa:
All disk read/write functions done through the CDBWrapper class in ClaimTrie should be separated out onto its own file/class. This should improve readability as it should be much clearer what is being saved and how. We should also write unit tests for the separated out disk read/write class.

bvbfan:
I will try to this task. You mean CDBWrapper (reads from disk) as well as CDBBatch (write to disk) in related class, right?

kaykurokawa:
Yes basically anything that has to do with CClaimTrie.db

bvbfan:
I found that architecture is not good enough to me. Did you accept more agressive refactor? Why, CClaimTrieCache should have all *dirty* nodes and CClaimTrie should search in, when it's not found, then query CClaimTrieDb (new separed class) so CClaimTrie became a interface between cache and db. This will result in remove of all dirtyQueueXXX and related functions from CClaimTrie. Is this acceptable to you?

bvbfan:
https://github.com/lbryio/lbrycrd/pull/140
It's simple, i can make more *aggressive* refactor if still needs

alyssaoc:
closing for #175

Issue #120, opened 2018-04-16
Travis build for OSX seems to fail occasionally ? 


kaykurokawa:
Getting these errors for the OSX Travis build: seems to be a new problem that started happening recently?  See : https://travis-ci.org/lbryio/lbrycrd/jobs/365289919 (also check current master)

==> Installing protobuf dependency: python@2
==> Downloading https://homebrew.bintray.com/bottles/python@2-2.7.14_3.el_capita
==> Pouring python@2-2.7.14_3.el_capitan.bottle.1.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-2
Target /usr/local/bin/2to3-2
is a symlink belonging to python. You can unlink it:
  brew unlink python
To force the link and overwrite all conflicting files:
  brew link --overwrite python@2
To list all files that would be deleted:
  brew link --overwrite --dry-run python@2

kaykurokawa:
The failure is consistently happening, not occasionally

lbrynaut:
Resolved by #165

PR #140, opened 2018-05-19
Separate disk related functions in CClaimTrieDb

bvbfan:

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,8 @@ If you encounter any errors, please check `doc/build-*.md` for further instructi
 
 Contributions to this project are welcome, encouraged, and compensated. For more details, see [lbry.io/faq/contributing](https://lbry.io/faq/contributing)
 
+The codebase is in C++03, C++11 is currently not supported but we will be migrating to it in the near future. Recommended GCC version is 4.8 or greater.
+
 The `master` branch is regularly built and tested, but is not guaranteed to be
 completely stable. [Releases](https://github.com/lbryio/lbrycrd/releases) are created
 regularly to indicate new official, stable release versions.
@@ -59,9 +61,7 @@ regularly to indicate new official, stable release versions.
 Testing and code review is the bottleneck for development; we get more pull
 requests than we can review and test on short notice. Please be patient and help out by testing
 other people's pull requests, and remember this is a security-critical project where any mistake might cost people
-lots of money.
-
-Developers are strongly encouraged to write [unit tests](/doc/unit-tests.md) for new code, and to
+lots of money. Developers are strongly encouraged to write [unit tests](/doc/unit-tests.md) for new code, and to
 submit new unit tests for old code. Unit tests can be compiled and run
 (assuming they weren't disabled in configure) with: `make check`
 


bvbfan:
@kaykurokawa, OSX build fails, but it's not looks like pull request related.

kaykurokawa:
Yes OSX build on travis is currently failing for all : https://github.com/lbryio/lbrycrd/issues/120

src/claimtrie.cpp
```diff
@@ -2,10 +2,24 @@
...
+#define TRIE_NODE 'n'
+#define CLAIM_BY_ID 'i'
+#define CLAIM_QUEUE_ROW 'r'
+#define CLAIM_QUEUE_NAME_ROW 'm'
+#define EXP_QUEUE_ROW 'e'
+#define SUPPORT 's'
+#define SUPPORT_QUEUE_ROW 'u'
+#define SUPPORT_QUEUE_NAME_ROW 'p'
+#define SUPPORT_EXP_QUEUE_ROW 'x'
+
```
kaykurokawa: I think these can go in claimtriedb.h

kaykurokawa:
Hi, as you mentioned in https://github.com/lbryio/lbrycrd/issues/136 , a bit more aggressive refactor would be acceptable. To be specific , the various dirty* memeber variables that is currently managed by CClaimTrie could be managed by CClaimTrieDb. This would make it much clearer exactly what kind of things are being written/read to disk and how, and should also improve some DRY problems  involved in the manipulation of dirty* member variables. 

Please note though I think there is a slight misunderstanding you may have about CClaimTrieCache , this class is NOT a cache of the CClaimTrie. CClaimTrieCache actually contains a cache of changes that must be applied to CClaimTrie upon block increment (removing claims, adding claims, etc..). Thus CClaimTrieCache should be minimally affected by this refactor.

bvbfan:
You're right about of CClaimTrieCache, i wrote the comment before i see the implementation, name lies to me, did you think it's proper name?

src/claimtrie.cpp
```diff
@@ -952,146 +938,58 @@ bool CClaimTrie::updateTakeoverHeight(const std::string& name, int nTakeoverHeig
...
+        WriteNode(itcache->first, itcache->second);
+
+    WriteQueue(CLAIM_QUEUE_ROW, dirtyQueueRows.begin(), dirtyQueueRows.end());
+    WriteQueue(CLAIM_QUEUE_NAME_ROW, dirtyQueueNameRows.begin(), dirtyQueueNameRows.end());
+    WriteQueue(EXP_QUEUE_ROW, dirtyExpirationQueueRows.begin(), dirtyExpirationQueueRows.end());
+    WriteQueue(SUPPORT, dirtySupportNodes.begin(), dirtySupportNodes.end());
+    WriteQueue(SUPPORT_QUEUE_ROW, dirtySupportQueueRows.begin(), dirtySupportQueueRows.end());
+    WriteQueue(SUPPORT_QUEUE_NAME_ROW, dirtySupportQueueNameRows.begin(), dirtySupportQueueNameRows.end());
+    WriteQueue(SUPPORT_EXP_QUEUE_ROW, dirtySupportExpirationQueueRows.begin(), dirtySupportExpirationQueueRows.end());
+
```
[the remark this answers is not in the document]
bvbfan: I'll investigate to move dirty queues as well as read/write in CClaimTrieDb

src/claimtrie.cpp
bvbfan: I've give a try to implement it on elegant way. Generic in claimtriedb as
```
template<typename K, typename V>
bool getQueue(const K &key, V &value) const
{
    auto hash = typeid(std::map<K, V>).hash_code();
    auto it = map_hashes.find(hash);
    if (it == map_hashes.end()) return false;
    auto map = reinterpret_cast<std::map<K, V>>(it->second);
    auto i = map.find(key);
    if (i != map.end()) { value = i->second; return true; }
    return false;
}
```
I mean it can be used a hash to store generic queues so all definitions will gone away, but we should provide db migration, right?

bvbfan:
I force push new version, but it looks like not updated here :)

bvbfan:
I can't figure out why when i write to leveldb CClaimTrieNode is not empty, but when i reads it is. Did you have any advice? Test fails about that
```
void CCMap<K, V>::write(size_t, CClaimTrieDb*) [with K = std::basic_string<char>; V = CClaimTrieNode; size_t = long unsigned int], , 0
void CCMap<K, V>::write(size_t, CClaimTrieDb*) [with K = std::basic_string<char>; V = CClaimTrieNode; size_t = long unsigned int], t, 0
void CCMap<K, V>::write(size_t, CClaimTrieDb*) [with K = std::basic_string<char>; V = CClaimTrieNode; size_t = long unsigned int], te, 0
void CCMap<K, V>::write(size_t, CClaimTrieDb*) [with K = std::basic_string<char>; V = CClaimTrieNode; size_t = long unsigned int], tes, 0
void CCMap<K, V>::write(size_t, CClaimTrieDb*) [with K = std::basic_string<char>; V = CClaimTrieNode; size_t = long unsigned int], test, 0
bool CClaimTrieDb::SeekFirstKey(K&, V&) const [with K = std::basic_string<char>; V = CClaimTrieNode], , 1
bool CClaimTrieDb::SeekFirstKey(K&, V&) const [with K = std::basic_string<char>; V = CClaimTrieNode], t, 1
bool CClaimTrieDb::SeekFirstKey(K&, V&) const [with K = std::basic_string<char>; V = CClaimTrieNode], te, 1
bool CClaimTrieDb::SeekFirstKey(K&, V&) const [with K = std::basic_string<char>; V = CClaimTrieNode], tes, 1
bool CClaimTrieDb::SeekFirstKey(K&, V&) const [with K = std::basic_string<char>; V = CClaimTrieNode], test, 0
```

bvbfan:
Ok, Kay, can review it? The architecture aims to provide queues by needs, for now they are maps. through CClaimTrieDb. I plan to refactor CClaimTrieChache too to benefit db cache and to be renamed, as we discuss. One Qt test still fail, somehow, i can't see this is my fault. Hope we can merge it in refactor branch after you explain your remarks.
Regards

bvbfan:
Did you have to time to take a look of this? After all i have one pending change, last implementation uses vector, under the hood, to not reorder cached queues.

kaykurokawa:
Will start looking at this today

bvbfan:
Kay i will change hash calculation since i check older version of boost, https://wandbox.org/ helped a lot, Boost.TypeIndex started at version 1.56, but the problem is around 1.64 it's change hash calculation, which is annoying. I will implement it in different way.

src/claimtrie.h
```diff
@@ -255,24 +241,77 @@ class CClaimIndexElement
...
     }
 
+    bool empty() const
+    {
+        return name.empty() && claim.claimId.IsNull();
+    }
+
     std::string name;
     CClaimValue claim;
 };
```
kaykurokawa: This block of code (253 - 293) is a bit confusing, it's basically just to enable swap() right? 
Maybe adding some comments about what its trying to achieve would be helpful.

src/claimtriedb.h
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Represents access to claims in database
+ */
+class CClaimTrieDb : public CDBWrapper
+{
+public:
+    CClaimTrieDb(bool fMemory = false, bool fWipe = false);
+
+    ~CClaimTrieDb();
+
```
kaykurokawa: Spelling : "durty" ->"dirty"

src/claimtrie.h
bvbfan: From 262 to 284 - yes, it for std::swap to work. Helpers and typedefs are to make a different type when insert to db, so let's see an example
```
typedef Generic<outPointHeightType, queueNameRowHelper> queueNameRowValueType;
typedef Generic<outPointHeightType, supportQueueNameRowHelper> supportQueueNameRowValueType;
```
Helper is used only to make these 2 types different e.g. to stored in different keys in db. I'll commented it.

src/claimtriedb.h
kaykurokawa: Actually, we should probably stop using this "dirty" term as it is not very descriptive. Maybe:
"Write to disk the buffered changes to the key value storage"

src/claimtriedb.h
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Writes all durty queues to disk
+     */
+    void writeQueues();
+
+    /**
+     * Gets a map representation of K type / V type stored by their hash
+     * @param[out] map  key / value pairs readed from queues and disk
+     */
```
kaykurokawa: "key / value pairs readed from queues and disk"  - > "key / value pairs read from disk with changes from the buffer applied"

src/claimtriedb.h
```diff
@@ -0,0 +1,130 @@
...
+    template <typename K, typename V, typename C>
+    bool getQueueMap(std::map<K, V, C> &map) const;
+
+    /**
+     * Gets a value of type V found by key of type K stored by their hash
+     * @param[in] key   key to looking for in dirty queues and disk
+     * @param[out] row  value which is found
+     */
+    template <typename K, typename V>
```
kaykurokawa: "key to looking for in dirty queues and disk" -> "key to look for in buffer and disk"

src/claimtriedb.h
```diff
@@ -0,0 +1,130 @@
...
+    bool getQueueRow(const K &key, V &row) const;
+
+    /**
+     * Update value of type V by key of type K through their hash
+     * @param[in] key       key to looking for in dirty queues and disk
+     * @param[in/out] row   update value and gets its last value
+     */
```
kaykurokawa: " key to looking for in dirty queues and disk" - > "key to look for in buffer and disk"

src/claimtriedb.h
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Check whatever exists values stored by hash
+     */
+    template <typename K, typename V>
+    bool keyTypeEmpty() const;
+
+    /**
+     * Get a map representation of K type / V type stored by theirs hash
+     * @param[out] map  key / value pairs readed only from disk
+     */
```
kaykurokawa: "key / value pairs readed only from disk" -> "key/value pairs, read only from disk"

src/claimtriedb.h
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Actual migation of key
+     * @param[in] old_key   old key to replace it
+     */
+    template <typename K, typename V>
+    void migrate(char old_key);
+
+    /**
+     * Represents dirty queues before stored to disk
+     */
```
kaykurokawa: "Represents buffer of changes"

src/claimtrie.h
bvbfan: line 253: /// Helpers to separate queue types from each other
line 271: /// Make std::swap to work with custom types
line 288: /// Each type will be stored in database separately

kaykurokawa: looks good

kaykurokawa:
Updated doc strings in 53cf80d99c2577582396612215b03bb1307b5a57 , 
See branch: seperate_disk_claim

bvbfan:
It looks good to me.

src/claimtrie.cpp
```diff
@@ -1130,91 +669,71 @@ bool CClaimTrie::ReadFromDisk(bool check)
...
-                else
-                {
-                    return error("%s(): error reading claim trie from disk", __func__);
-                }
-            }
+    trieNodeMapType nodes;
+    if(!db.seekByKey(nodes)) {
+        return error("%s(): error reading claim trie from disk", __func__);
+    }
+
```
kaykurokawa: Upon fresh start (with empty data directory), there won't be any nodes to load, so this will cause lbrycrdd to exit. See
https://github.com/lbryio/lbrycrd/blob/master/src/init.cpp#L1286

kaykurokawa:
I found a problem, try syncing lbrycrdd from scratch (clear out data directory before starting lbrycrdd).
See comment above, should be a simple fix.

bvbfan:
```
diff --git a/src/claimtriedb.cpp b/src/claimtriedb.cpp
index 0c0eccf7..198d6641 100644
--- a/src/claimtriedb.cpp
+++ b/src/claimtriedb.cpp
@@ -136,20 +136,17 @@ bool CClaimTrieDb::seekByKey(std::map<K, V, C> &map) const
     const size_t hash = hashType<K, V>();
     boost::scoped_ptr<CDBIterator> pcursor(const_cast<CClaimTrieDb*>(this)->NewIterator());
 
-    bool found = false;
-
     for (pcursor->SeekToFirst(); pcursor->Valid(); pcursor->Next()) {
         std::pair<size_t, K> key;
         if (pcursor->GetKey(key)) {
             if (hash == key.first) {
                 V value;
                 if (!pcursor->GetValue(value)) return false;
-                found = true;
                 map.insert(std::make_pair(key.second, value));
             }
         }
     }
-    return found;
+    return true;
 }
 
 template <typename K, typename V, typename C>
```

bvbfan:
So maybe we should update test as well
claimtriebranching_tests.cpp:319 
```
BOOST_CHECK(pclaimTrie->ReadFromDisk(true));
```

kaykurokawa:
Yes, these changes look good, also add a docstring to seekByKey() 

"Returns false if database read fails."

bvbfan:
Also notice that i have pending changes to that branch
https://github.com/lbryio/lbrycrd/issues/145

kaykurokawa:
Closing as further work is continued in #160

PR #142, opened 2018-05-21
Add support information to getclaimbyid rpc

### LINKED ISSUE #124: Add supports on getclaimbyid RPC command
`getclaimbyid` calculates the effective amount, but doesn't returns the supports list. Supports are needed for our electrum layer. For getting it we actually need to call `getclaimsforname`, which transfer more data then needed and adds an extra call.

@lbrynaut proposed a new index for supports over claim ids. I think it's the way to go, but we should discuss here.

I have some very ugly proof of concept here with a new JSON format for it: https://github.com/shyba/lbrycrd/commit/2fdd202cbf0908ca9441e6edbd6e667ce5c9a7a1

Some other idea was to make the support listing optional with an extra boolean parameter like `includeSupports`.

lbrynaut:
Resolves #124


src/claimtrie.cpp
```diff
@@ -525,31 +525,41 @@ claimsForNameType CClaimTrie::getClaimsForName(const std::string& name) const
     return allClaims;
 }
 
-//return effective amount form claim, retuns 0 if claim is not found
+//return effective amount from claim, retuns 0 if claim is not found
```
shyba: good catch

src/rpc/claimtrie.cpp
```diff
@@ -387,9 +399,20 @@ UniValue getclaimbyid(const UniValue& params, bool fHelp)
...
+            supportEntry.push_back(Pair("txid", support.outPoint.hash.GetHex()));
+            supportEntry.push_back(Pair("n", (int)support.outPoint.n));
+            supportEntry.push_back(Pair("height", support.nHeight));
+            supportEntry.push_back(Pair("valid at height", support.nValidAtHeight));
+            supportEntry.push_back(Pair("amount", support.nAmount));
+            supportList.push_back(supportEntry);
+        }
+        claim.push_back(Pair("supports", supportList));
         claim.push_back(Pair("height", claimValue.nHeight));
+        claim.push_back(Pair("valid at height", claimValue.nValidAtHeight));
```
kaykurokawa: "valid at height" field was added here but not to the doc string

lbrynaut: @kaykurokawa Which docstring? It was added to the help string.

kaykurokawa: its added when listing fields for supports, but not to the claim itself in the docstring

kaykurokawa:
untested but LGTM

src/rpc/claimtrie.cpp
lbrynaut: Ah, I see, thanks!

lbrynaut: Updated

PR #146, opened 2018-05-24
fix fPowAllowMinDifficultyBlocks 

kaykurokawa:
We need to make sure that the new fPowAllowMinDifficultyBlocks only affects block 277299 and onwards, otherwise previous blocks that were created without this logic will be invalid because the difficulty calculations for the block will be incorrect.

### DIFF
--- a/src/pow.cpp
+++ b/src/pow.cpp
@@ -18,13 +18,14 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
     if (pindexLast == NULL)
         return nProofOfWorkLimit;
 
-    if (params.fPowAllowMinDifficultyBlocks)
+    if (params.fPowAllowMinDifficultyBlocks && pindexLast->nHeight >= 277299)
     {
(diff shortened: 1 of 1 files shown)


src/pow.cpp
```diff
@@ -18,13 +18,14 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
     if (pindexLast == NULL)
         return nProofOfWorkLimit;
 
-    if (params.fPowAllowMinDifficultyBlocks)
```
shyba: IMHO without the height check it made sense to simply check for `fPowAllowMinDifficultyBlocks`, but now that there is a hardcoded height I think it would make sense to explicitly check testnet instead of fPowAllowMinDifficultyBlocks since that's the real case of the condition right?

shyba:
LGTM. I made a comment for a possible small refactor to clarify the if clause but then I saw the comment block was also updated and very clear, so I'm ok with it.

PR #148, opened 2018-05-26
Fix newest boost and GCC warnings

bvbfan:

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3081,7 +3081,7 @@ UniValue fundrawtransaction(const UniValue& params, bool fHelp)
     if (origTx.vout.size() == 0)
         throw JSONRPCError(RPC_INVALID_PARAMETER, "TX must have at least one output");
 
-    if (changePosition != -1 && (changePosition < 0 || changePosition > origTx.vout.size()))
+    if (changePosition != -1 && (changePosition < 0 || static_cast<size_t>(changePosition) > origTx.vout.size()))
         throw JSONRPCError(RPC_INVALID_PARAMETER, "changePosition out of bounds");
 
     CMutableTransaction tx(origTx);
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -404,21 +404,21 @@ bool CWallet::Verify()
         } catch (const boost::filesystem::filesystem_error&) {
             // failure is ok (well, not really, but it's not worse than what we started with)
         }
-        
+
         // try again
         if (!bitdb.Open(GetDataDir())) {
             // if it still fails, it probably means we can't even create the database env
             return UIError(strprintf(_("Error initializing wallet database environment %s!"), GetDataDir()));
         }
     }
-    
+
     if (GetBoolArg("-salvagewallet", false))
     {
         // Recover readable keypairs:
         if (!CWalletDB::Recover(bitdb, walletFile, true))
             return false;
     }
-    
+
     if (boost::filesystem::exists(GetDataDir() / walletFile))
     {
         CDBEnv::VerifyResult r = bitdb.Verify(walletFile, CWalletDB::Recover);
@@ -433,7 +433,7 @@ bool CWallet::Verify()
         if (r == CDBEnv::RECOVER_FAIL)
             return UIError(strprintf(_("%s corrupt, salvage failed"), walletFile));
     }
-    
+
     return true;
 }
 
@@ -1762,9 +1762,9 @@ static void ApproximateBestSubset(vector<pair<CAmount, pair<const CWalletTx*,uns
         }
     }
 
-    //Reduces the approximate best subset by removing any inputs that are smaller than the surplus of nTotal beyond nTargetValue. 
+    //Reduces the approximate best subset by removing any inputs that are smaller than the surplus of nTotal beyond nTargetValue.
     for (unsigned int i = 0; i < vValue.size(); i++)
-    {                        
+    {
         if (vfBest[i] && (nBest - vValue[i].first) >= nTargetValue )
         {
             vfBest[i] = false;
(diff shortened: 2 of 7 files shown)


src/wallet/wallet.cpp
```diff
@@ -2198,12 +2198,12 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
...
                     else
                     {
-                        if (nChangePosInOut == -1)
+                        if (nChangePosInOut < 0)
                         {
                             // Insert change txn at random position:
                             nChangePosInOut = GetRandInt(txNew.vout.size()+1);
                         }
-                        else if (nChangePosInOut > txNew.vout.size())
+                        else if (size_t(nChangePosInOut) > txNew.vout.size())
```
lbrynaut: Use static_cast

src/wallet/wallet.cpp
```diff
@@ -2198,12 +2198,12 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
                     }
                     else
                     {
-                        if (nChangePosInOut == -1)
+                        if (nChangePosInOut < 0)
```
lbrynaut: Why change this?

src/wallet/rpcwallet.cpp
```diff
@@ -3081,7 +3081,7 @@ UniValue fundrawtransaction(const UniValue& params, bool fHelp)
     if (origTx.vout.size() == 0)
         throw JSONRPCError(RPC_INVALID_PARAMETER, "TX must have at least one output");
 
-    if (changePosition != -1 && (changePosition < 0 || changePosition > origTx.vout.size()))
+    if (changePosition != -1 && (changePosition < 0 || size_t(changePosition) > origTx.vout.size()))
```
lbrynaut: Use static_cast

src/wallet/wallet.cpp
lbrynaut: The cast is for the unsigned comparison(s), not changing the type.

bvbfan: I mean static_cast<size_t>(-2) will result in insert container past end. line: 2206

bvbfan: Ah, forgot it. Check is mistaken.

src/httpserver.cpp
```diff
@@ -252,7 +252,7 @@ static std::string RequestMethodString(HTTPRequest::RequestMethod m)
 /** HTTP request callback */
 static void http_request_cb(struct evhttp_request* req, void* arg)
 {
-    std::auto_ptr<HTTPRequest> hreq(new HTTPRequest(req));
+    std::unique_ptr<HTTPRequest> hreq(new HTTPRequest(req));
```
lbrynaut: We're not yet using C++11.  Use boost unique ptr in the meantime.

src/httpserver.cpp
```diff
@@ -288,7 +288,7 @@ static void http_request_cb(struct evhttp_request* req, void* arg)
 
     // Dispatch to worker thread
     if (i != iend) {
-        std::auto_ptr<HTTPWorkItem> item(new HTTPWorkItem(hreq.release(), path, i->handler));
+        std::unique_ptr<HTTPWorkItem> item(new HTTPWorkItem(hreq.release(), path, i->handler));
```
lbrynaut: Same as above.

src/httpserver.cpp
bvbfan: We can't use it. The checks in configure are >= 1.20, unique_ptr lands in 1.57 i'll update patch.

lbrynaut:
@bvbfan You're right, the required boost is a bit dated.  Rather than change that, I think the slight performance hit of using boost::shared_ptr instead is the way to go instead of unique_ptr.

bvbfan:
Furthermore i will return auto_ptr, since we can plan to switch for C++11, hopefully soon :) I compile project with C++11 and did not realize that is not C++11 :)

lbrynaut:
@bvbfan Right, not C++11.  The claimtrie could be completely re-factored and cleaned up when we support C++11, but alas, it remains where it is for now.  It's on the roadmap (at least we've discussed it before).

bvbfan:
Yes, i made a huge refactor of CClaimTrie not only to move db in separate class / file even CClaimCache should be refactored and renamed, as discussed before.

lbrynaut:
@bvbfan I'm aware.  I meant a C++11 mechanical language re-factor not an organizational/architectural re-factor.

bvbfan:
Yes, i hope after architectural refactor, step to C++11 refactor will be easiest, i hope, second time :)

bvbfan:
You can merge this, if you does not have any other remarks. :)

lbrynaut:
It looks good to me.  I'm waiting on any input from @kaykurokawa, otherwise may merge later this afternoon.

kaykurokawa:
I'm a bit hesitant to merge this.  If we are modifying Bitcoin Core code like this, than there will be a lot of fixes, and we will have a bad time trying to merge from upstream which we will eventually have to do. 

I checked through some of these changes in the most recent bitcoin code base and it does indeed seem to be fixed  in the same way you did. Maybe the better way to fix it would be to cherry-pick these commits from Bitcoin Core (not saying you have to do this for this PR, but in the future, I would check to see if you aren't redoing what has already been done). 

It is important though to fix warnings for our quality of life purposes. I think we should get a process in place to say that "hey this commit can be ignored once we merge upstream", this way we can just remove ignore them and merging will be easier. Maybe something like putting in the commit comments, "NOT FOR MERGE UPSTREAM" ? We can just search and remove these commits. Or maybe these types of changes all go in to the same branch and  get merged from that branch?

bvbfan:
I don't notice that they changed upstream, it can be cherry-pick from upstream Core for master, then ignored when merge it from my specific branch, it is OK?

lbrynaut:
@kaykurokawa Good point.  I say we punt on this until we formally merge from upstream.  It's something that I plan to start in the next month or two, so I think it's better to not make things potentially harder later at this point.

tzarebczan:
What's the status of this pr? @bvbfan

lbrynaut:
@tzarebczan I'm of the opinion that it was a good exercise, but should be closed unmerged for now.  Upstream changes being pulled in are on the roadmap, which should resolve a lot of this.

lyoshenka:
If we cherry-pick commits from upstream, I don't think we need to mark them in any special way. Git should be able to handle this, and either way Kay told me its likely the way we get up to date is by starting with the current Core code and merging in our stuff. So maybe there's no harm in merging this?

@kaykurokawa correct me if I'm wrong.

bvbfan:
So this should be closed i think

PR #149, opened 2018-05-27
docker: Lbrynet as a container

chamunks:
## Documentation 

Added some basic documentation to docker/README.md it can be used elsewhere but at some point if we add other things which end up justifying that the end user run a `git clone` and then `docker-compose up -d` from inside of the git repo of lbrycrd we can let them store data in docker/data/ with some fancy git ignoring.

## Docker Compose
For the sake of getting people launched asap I've included a basic docker-compose.yml file which can launch the app with `docker-compose up -d` then commands can be run using `docker exec CONTAINERNAME lbrynet-cli commands`

## .gitlab-ci.yml
This is for CI/CD prebuild and deploy if we want to use GitLab for building these containers rather than the hub.docker.com route.  Currently, this is just configured to show what it would look like with my repositories.

## Dockerfile

Currently, the dockerfile is very basic and just installs the daemon into the containers $PATH for easy execution then it stores the daemons content in /app/ as that is defined as the unprivileged daemon processes home directory. 

## Commit size
https://en.wikipedia.org/wiki/Atomic_commit#Atomic_commit_convention the version I was working from in my topic branch was hugely out of date so rather than carrying over all of the commits I just added everything fresh I know that it's lazy but this code doesn't meddle with anyone's stuff so it seemed like it should be okay to do it this way.

## Maintainance

This container should be reasonably future proof for now as I've defined all non-ephemeral blobs to reside in the /app/ directory inside of the container which can be updated at a later time just fine.

The configuration will be added using a shell script in the future which can be optionally included.  The goal is to allow configuration using environment variables in the docker-compose.yml file which will be the officially supported launch method as it's the least maintenance method.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,14 @@
+image: ubuntu:18.04
+
+before_script:
+  - docker login -u gitlab-ci-token -p $CI_JOB_TOKEN $CI_REGISTRY
+
+build:
+  stage: build
+  script:
+  ## Add a prep.sh file eventually when we add a configuration function to this container.
+  # - ./docker/prep.sh
+  - docker build -t registry.gitlab.com/chamunks/lbry/$CI_COMMIT_REF_NAME:latest .
+  - docker-compose stop
+  - docker-compose rm -f
+  - docker-compose up -d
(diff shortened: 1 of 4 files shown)


chamunks:
Ignore this I added these files to the wrong repo for the PR.  I've got an annoying setup so that I can do CI/CD with GitLab as I don't have access to any of the LBRY testing mechanisms like Travis.

PR #150, opened 2018-05-28
cleanup: revert unused changes

roylee17:
to reduce diff from upstream.


roylee17:
never mind... it was use in CheckProofOfWork2

PR #102, opened 2018-02-28
Claim name normalization

lbrynaut:
Add lowercase and unicode normalization to claim name entries in the claim trie

Related to #65

Issue #106, opened 2018-03-14
Reimplement ClaimTrie as Patricia Trie


lyoshenka:
Our Claimtrie is inefficient because it must store lots of empty nodes when storing a long name. The solution is to use this: https://github.com/ethereum/wiki/wiki/Patricia-Tree

shyba:
For the sake of having this documented, we should be mindful that there is a space efficiency optimization and a time efficiency one mixed in this change. If we are doing one or both is a pending discussion.
If we just change the trie without changing the how the hash is calculated we could solve the memory issue without a hardfork. While changing how the proofs are made to match the new data structure (skipping empty nodes) would require a hardfork.

BrannonKing:
I was just doing some testing on node counts for this. With a million random strings (standard 60 characters), I end up with 25 million nodes in traditional trie. With the collapsed prefixes, I end up with 1.5 million nodes after a million insertions. That's a substantial difference! Next test: the qp trie. We can talk about where to check in my tests and what to name them next week.

BrannonKing:
I'm attaching my experiments on this. I've concluded that the prefix trie is the right structure, and that it would drastically reduce our (present) memory usage. I've also decided that it won't be enough code to justify putting it in a separate library by itself.  I'll leave it here on the shelf and hope that we can get c++11 support in lbrycrd in the very near future.

[lbry_tries.zip](https://github.com/lbryio/lbrycrd/files/2394602/lbry_tries.zip)

BrannonKing:
Last week I tested the code here with my same sample data: https://github.com/ytakano/radix_tree. It was slow and seemed to be a few nodes short of a full load. I didn't see a lot of other libraries that were easily accessible. I did not attempt to use Ethereum's trie code. It appeared complex, and I don't want to mix their code in with ours.

Issue #65, opened 2017-06-30
Case-insensitive claims


kauffj:
LBRY URLs ought to be case-insensitive, the way traditional domains are. Two claims with names with different case-sensitivity should be considered having the same name.

aayanl:
How would this affect already existing claims?

kaykurokawa:
a) We need to have the claimtrie work with unicode, not bytes if we are attaching strict interpretations of what the bytes means, are non unicode bytes now invalid? Which unicode encoding are we using? 

a) We need to define "case" here. English is of course easy, but it is kind of silly to deal with just English "cases" but ignore other languages. Is there a standard we can adopt or work off of here?  

c) as aayani said, we need to determine what happens to the already existing claims when this change occurs. (Does claim dog and DOG get merged into the same name?) and implement this logic. 

It is also possible to do this via soft fork by rejecting claims with cases in them after a certain block and letting already existing ones expire.

kauffj:
a) I'm not sure of the answer here, but if some claims get invalidated that would acceptable (though obviously not ideal). I don't have a problem with supporting non-Unicode or arbitrary byte claims, so long as when/if the claim is valid Unicode, we are considering case-sensitivity. 

b) I believe Unicode already defines lower and upper-case characters, as you can use regular expressions categories like `\p{Ll}` and `\p{Lu}` to refer to upper and lower case characters.

c) Yes, these would get merged into the same name under this proposal.

I'd prefer the solution here to be that claims retain the original provided capitalization, rather than preventing capitalization in claims. That is, we want someone to be able to claim `lbry://eXistenZ` and retain the capitalization, but still have `lbry://existenz` resolve to it.

tzarebczan:
So if two people own the same channel/claim name with different capitalizations, the resolve will default to the one with the higher bid right?

kauffj:
Yes, that's the idea.

etisdew:
What is the status of this feature? I'd like to counter that by the argument of address space being tightly coupled to length of string and variance in characters. This added expense to hashing is equivalent to mandating everything be in lower-case and have a doubly long url.

kaykurokawa:
This provides a standarized method of normalization for unicode characters: http://unicode.org/reports/tr15/#Notation , and is currently being proposed as the method for implementation case-insensitve claims along with other normalizations.

kaykurokawa:
Here is a unicdoe vulnerability that was floating around as a phishing attack on Telegram recently: 
[...]

PR #159, opened 2018-06-12
Implement a hard fork to normalize claim names

lbrynaut:
This is a complete re-write of #102 and replaces it since it's much more complete and handles situations that the other does not handle.

This PR contains a lot of changes and I expect a detailed review will take some time to ensure correctness.  I also expect the reproducible_build script will need further modification (likely won't work as-is again due to the added ICU dependency).

@kaykurokawa The "claimtriebranching_hardfork_disktest" unit test does not seem to work with this PR, if you have a second to see what's going on there, it would help. It came up after things were working and then I rebased to latest, so it's commented out for now.

EDIT: @kaykurokawa This last comment is no longer true, so can be ignored now.


lbrynaut:
Addresses #65 

WIP regarding travis/dependency addition.

configure.ac
```diff
@@ -200,7 +206,8 @@ fi
 ##       compatibility with the legacy buildsystem.
 ##
 if test "x$CXXFLAGS_overridden" = "xno"; then
-  CXXFLAGS="$CXXFLAGS -Wall -Wextra -Wformat -Wformat-security -Wno-unused-parameter -Wno-self-assign"
+#  CXXFLAGS="$CXXFLAGS -Wall -Wextra -Wformat -Wformat-security -Wno-unused-parameter -Wno-self-assign"
```
BrannonKing: Delete it if it's obsolete.

src/claimtrie.cpp
```diff
@@ -578,7 +677,10 @@ bool CClaimTrie::recursiveCheckConsistency(const CClaimTrieNode* node) const
     {
         if (recursiveCheckConsistency(it->second))
         {
-            vchToHash.push_back(it->first);
+            std::stringstream ss;
```
BrannonKing: This exists just to convert a char to a string?

packaging/build_windows.sh
```diff
@@ -6,13 +6,32 @@ sudo apt-get update
...
+     libssl-dev libevent-dev bsdmainutils curl ca-certificates libicu-dev
+
+#################################################################
+# Build ICU for Linux first so that we can cross compile it below
+#################################################################
+staging_dir=/tmp/icu_staging
+icu_linux_dir=$staging_dir/build_icu_linux
+mkdir -p $staging_dir
+pushd $staging_dir
+wget http://download.icu-project.org/files/icu4c/55.1/icu4c-55_1-src.tgz
```
BrannonKing: boost.locale does not require this? Why would we need to get this separately from boost?

src/claimtrie.cpp
```diff
@@ -137,6 +150,77 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
     std::make_heap(claims.begin(), claims.end());
 }
 
+CClaimTrie* CClaimTrie::enableNormalizationFork(bool enable, int nHeight)
+{
+    LogPrintf("%s: Normalize Name Fork is %s\n",
+              __func__, (enable ? "ENABLED" : "DISABLED"));
+
+    if (enable)
```
BrannonKing: We default this to true after testing but before the release?

src/claimtrie.cpp
lbrynaut: A normalized string, yes.

packaging/build_windows.sh
lbrynaut: Because we're using icu for normalization, which doesn't work through boost unless it's explicitly enabled.

src/claimtrie.cpp
lbrynaut: I don't know what you're asking.  It's a flag specified by a caller.

packaging/build_windows.sh
BrannonKing: So what is the gain with using boost's wrapper? Suppose we just make ICU calls directly?

lbrynaut: It's C++ and style-compatible.  I also have previous experience using it, for exactly this kind of purpose.

src/claimtrie.h
```diff
@@ -28,6 +34,25 @@
...
+// lower-case and normalize any input string name
+// see: https://unicode.org/reports/tr15/#Norm_Forms
+inline const std::string normalizeClaimName(const std::string& name)
+{
+    boost::locale::localization_backend_manager manager =
+        boost::locale::localization_backend_manager::global();
+    manager.select("icu");
+
+    boost::locale::generator curLocale(manager);
+    const std::locale utf8 = curLocale("en_US.UTF8");
```
BrannonKing: It appears that the `utf8` object is immutable. Can we move the construction of it out of our inline method? It seems we could just construct a static instance once and be done with it.

lbrynaut:
Updated.

packaging/build_windows.sh
```diff
@@ -6,13 +6,32 @@ sudo apt-get update
...
+popd
 
 cd depends
 make HOST=i686-w64-mingw32 NO_QT=1
 cd ..
 patch -p1 < packaging/remove_consensus.patch
 ./autogen.sh
-./configure --prefix=`pwd`/depends/i686-w64-mingw32 --without-gui
+icu_mingw_dir=$(cat /tmp/icu_install_dir)
+echo "Using ICU MinGW Install dir $icu_mingw_dir"
```
lbrynaut: Remove echo

reproducible_build.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -euo pipefail
+set -xeuo pipefail
```
lbrynaut: Remove step output

reproducible_build.sh
```diff
@@ -27,7 +27,7 @@ function HELP {
 }
 
 CLONE=true
-CLEAN=false
+CLEAN=true
```
lbrynaut: Revert

reproducible_build.sh
```diff
@@ -106,24 +106,22 @@ else
     # this file is created when the build starts
     START_TIME_FILE="$TRAVIS_BUILD_DIR/start_time"
 fi
-if [ ! -f "${START_TIME_FILE}" ]; then
-    date +%s > "${START_TIME_FILE}"
-fi
-
+rm -f ${START_TIME_FILE}
+date +%s > ${START_TIME_FILE}
```
lbrynaut: Note: This makes the build times accurate

reproducible_build.sh
```diff
@@ -244,8 +242,10 @@ function install_apt_packages() {
 
 function build_dependencies() {
     if [ "${OS_NAME}" = "osx" ]; then
+        PARALLEL="-j $(sysctl -n hw.ncpu)"
 	install_brew_packages
     else
+        PARALLEL="-j $(grep -c processor /proc/cpuinfo)"
```
lbrynaut: Note that travis builds will be parallelized by this

src/chainparams.cpp
```diff
@@ -136,6 +136,7 @@ class CMainParams : public CChainParams {
         consensus.nExtendedClaimExpirationForkHeight = 400155;
         consensus.fPowAllowMinDifficultyBlocks = false;
         consensus.fPowNoRetargeting = false;
+        consensus.nNormalizedNameForkHeight = 1000000; // FIXME: pick a real fork height
```
lbrynaut: All fork heights are deferred to @kaykurokawa

src/chainparams.cpp
```diff
@@ -223,6 +224,7 @@ class CTestNetParams : public CChainParams {
         consensus.nExtendedClaimExpirationForkHeight = 278160;
         consensus.fPowAllowMinDifficultyBlocks = true;
         consensus.fPowNoRetargeting = false;
+        consensus.nNormalizedNameForkHeight = 1000000; // FIXME: pick a real fork height
```
lbrynaut: All fork heights are deferred to @kaykurokawa

src/chainparams.cpp
```diff
@@ -303,6 +305,7 @@ class CRegTestParams : public CChainParams {
         consensus.nExtendedClaimExpirationForkHeight = 800;
         consensus.fPowAllowMinDifficultyBlocks = false;
         consensus.fPowNoRetargeting = false;
+        consensus.nNormalizedNameForkHeight = 2000; // FIXME: pick a real fork height
```
lbrynaut: All fork heights are deferred to @kaykurokawa

src/claimtrie.cpp
```diff
@@ -137,6 +150,77 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
...
+    {
+        std::vector<namedNodeType> nodes = flattenTrie();
+        LogPrintf("Pref-normalization fork claim trie nodes = %lu\n", nodes.size());
+        // Debug-only
```
lbrynaut: Remove debug

src/claimtrie.cpp
```diff
@@ -137,6 +150,77 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
...
+            LogPrintf("Post-normalization fork claim trie nodes = %lu\n", nodes.size());
+            // Debug-only
+            /* for (std::vector<namedNodeType>::iterator it = nodes.begin(); it != nodes.end(); ++it) */
+            /* { */
+            /*     if (!it->second.claims.empty()) */
+            /*         LogPrintf("[***] Claim name: \"%s\" (hash %s)\n", it->first.c_str(), it->second.hash.GetHex().c_str()); */
+            /*     else */
+            /*         LogPrintf("[EMPTY] Claim name: \"%s\" (hash %s)\n", it->first.c_str(), it->second.hash.GetHex().c_str()); */
+            /* } */
+        }
```
lbrynaut: Remove debug

src/claimtrie.cpp
lbrynaut: To be clear, this fork implements normalization by removing the concept of a claimtrie of _chars_ and we instead have a claimtrie of _code points_, represented as chars (via utf8).

src/claimtrie.cpp
```diff
@@ -1222,36 +1364,35 @@ bool CClaimTrieCache::recursiveComputeMerkleHash(CClaimTrieNode* tnCurrent, std:
     }
 
     CHash256 hasher;
-    std::vector<unsigned char> vchHash(hasher.OUTPUT_SIZE);
     hasher.Write(vchToHash.data(), vchToHash.size());
+    std::vector<unsigned char> vchHash(hasher.OUTPUT_SIZE);
     hasher.Finalize(&(vchHash[0]));
+
```
lbrynaut: Style: remove blank lines

src/claimtrie.cpp
```diff
@@ -1331,26 +1462,22 @@ bool CClaimTrieCache::insertClaimIntoTrie(const std::string& name, CClaimValue c
         // order to calculate the merkle hash.
         cachedNode = cache.find(sCurrentSubstring);
         if (cachedNode != cache.end())
-        {
```
lbrynaut: Style revert unnecessary changes

src/test/claimtrie_tests.cpp
```diff
@@ -205,6 +205,7 @@ BOOST_AUTO_TEST_CASE(claimtrie_merkle_hash)
 
     BOOST_CHECK(pclaimTrie->empty());
     BOOST_CHECK(!ntState.empty());
+    fprintf(stderr, "%s: %s != %s\n", __func__, ntState.getMerkleHash().GetHex().c_str(), hash1.GetHex().c_str());
```
lbrynaut: Remove fprintf

src/test/claimtriebranching_tests.cpp
```diff
@@ -999,69 +1000,178 @@ BOOST_AUTO_TEST_CASE(claimtriebranching_hardfork_support)
 
 }
 
+/* /\* */
+/*     claim/support expiration for hard fork, but with checks for disk procedures */
+/* *\/ */
+/* BOOST_AUTO_TEST_CASE(claimtriebranching_hardfork_disktest) */
```
lbrynaut: @kaykurokawa Let's talk about this test at some point.

src/claimtrie.cpp
```diff
@@ -137,6 +149,75 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
...
+    for (queueNameType::const_iterator it = other.dirtyQueueNameRows.begin(); it != other.dirtyQueueNameRows.end(); ++it)
+        dirtyQueueNameRows[normalizeClaimName(it->first)] = it->second;
+    for (expirationQueueType::const_iterator it = other.dirtyExpirationQueueRows.begin(); it != other.dirtyExpirationQueueRows.end(); ++it)
+        dirtyExpirationQueueRows[it->first] = it->second;
+    for (supportQueueType::const_iterator it = other.dirtySupportQueueRows.begin(); it != other.dirtySupportQueueRows.end(); ++it)
+        dirtySupportQueueRows[it->first] = it->second;
+    for (expirationQueueType::const_iterator it = other.dirtySupportExpirationQueueRows.begin(); it != other.dirtySupportExpirationQueueRows.end(); ++it)
+        dirtySupportExpirationQueueRows[it->first] = it->second;
+    for (nodeCacheType::const_iterator it = other.dirtyNodes.begin(); it != other.dirtyNodes.end(); ++it)
+        dirtyNodes[normalizeClaimName(it->first)] = new CClaimTrieNode();
```
bvbfan: Does we need to copy trie node?
```
dirtyNodes[normalizeClaimName(it->first)] = new CClaimTrieNode(*(it->second));
```

src/claimtrie.h
```diff
@@ -28,6 +34,30 @@
 
 uint256 getValueHash(COutPoint outPoint, int nHeightOfLastTakeover);
 
+// lower-case and normalize any input string name
+// see: https://unicode.org/reports/tr15/#Norm_Forms
+inline const std::string normalizeClaimName(const std::string& name)
```
bvbfan: Returning const on non reference is non-sense in C++

kaykurokawa:
I opened up a separate issue to discuss the issue of "how to do character normalization" here: https://github.com/lbryio/lbrycrd/issues/204 .

Seems like the method in this PR may not be optimal but we can discuss that separately since I think it will not affect the bulk of the changes here.

BrannonKing:
My first attempt at compiling this branch gave me this error:
```
../libtool: line 7485: cd: auto/lib: No such file or directory
libtool:   error: cannot determine absolute directory name of 'auto/lib'
```
I am using system-installed boost (with the OpenSSL 1.1 PR) as I haven't been able to make the reproducible_build.sh work for me since July (on Bionic). After getting the above error, I validated that "auto" was the path for ICU, googled a bit, and read the configure.ac a bit. I decided to add this to the configure command:
```
--with-icu=`pkg-config icu-i18n --variable=prefix`
```
That did alleviate the issue. It seems that configure.ac is not correctly falling back to pkgconfig in the situation where `--with-icu` is omitted.

BrannonKing:
Another issue I ran into: I can query the claimtrie while it is being rebuilt in normalized form -- I get partial results. I think I should get some error instead, some "still loading" error like it gives me if I query during init.

src/claimtrie.h
lbrynaut: I'm keeping inline in place of a storage specifier though, will remove the const.

lbrynaut:
@BrannonKing reported a performance issue, and it's true, this is slower than it needs to be.  Correctness first, then performance is how I've always operated ;-)

Anyway, I've already got a partial fix, but will be updating when I think it's ready.

src/claimtrie.cpp
bvbfan: But why we want an empty node, it can wipe db in some cases.

src/claimtrie.cpp
```diff
@@ -137,6 +149,75 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
...
+        /* { */
+        /*     if (!it->second.claims.empty()) */
+        /*         LogPrintf("[---] Claim name: \"%s\" (hash %s)\n", it->first.c_str(), it->second.hash.GetHex().c_str()); */
+        /*     else */
+        /*         LogPrintf("[EMPTY] Claim name: \"%s\" (hash %s)\n", it->first.c_str(), it->second.hash.GetHex().c_str()); */
+        /* } */
+
+        if (normalizationForkStateInstance.nonNormalizedTrie == NULL) {
+            normalizationForkStateInstance.nonNormalizedTrie = this;
+            normalizationForkStateInstance.normalizedTrie = new CClaimTrie(false, false, true);
```
bvbfan: Leaking this is not a problem right now, but if we use placement or in-class operator new (that i've plans to do it) will be.
```
static CClaimTrie normalizedTrie(false, false, true);
normalizationForkStateInstance.normalizedTrie = &normalizedTrie;
```
It'll be better

BrannonKing:
My experience with this branch thus far (with fork height at 1M):
1. I originally ran it with my existing mainnet data (including txindex). I was told that I needed to reindex.
2. I let the reindex run over the weekend; it was soon obvious that it needed a lot of time. It took 50+ hours. 
3. I stopped the server so I could set the fork block height, recompiled, and reran. I was again told that I needed to reindex. 
4. I then fired up a lbrycrdd build from a different branch. It crashed with this error: `main.cpp:2124: bool DisconnectBlock(const CBlock&, CValidationState&, const CBlockIndex*, CCoinsViewCache&, CClaimTrieCache&, bool*): Assertion 'pindex->GetBlockHash() == trieCache.getBestBlock()' failed.`

src/claimtrie.cpp
bvbfan: Mine concern is that operator delete isn't called, it can do many other things than to free memory.

lbrynaut: Understood, but it needs to remain in memory and intact past the fork.  A future release will remove it (by deleting it) when it's clear that we can.  If you're generally pointing out that the single pointer is leaked, it's not particularly helpful because we know there are leaks in the claimtrie already (and this one isn't one in particular to be concerned about, IMO).  The proposed suggestion doesn't change that either.  Upstream uses managed pointers, and while our first update to it will not, it will be much easier to transition after that update.  I was originally planning to do that before the upstream release, but it makes sense to do it after since we can skip boost pointers altogether and just use std ones.

src/claimtrie.cpp
lbrynaut: Normalization can narrow the claimtrie name space.

lbrynaut:
> My experience with this branch thus far (with fork height at 1M):

@BrannonKing I dusted off the webs, please test and find more (fork related) issues.

BrannonKing:
The reindex issue is resolved. However, I was unable to run past the fork height. I modified the fork height to occur in the next few minutes. However, it never went past the block height one before my fork height. I waited about 20min. This is on mainnet.

src/claimtrie.h
```diff
@@ -28,6 +34,34 @@
...
+
+        static boost::locale::generator curLocale(manager);
+        utf8 = curLocale("en_US.UTF8");
+        initialized = true;
+    }
+
+    std::string normalized = boost::locale::normalize(name, boost::locale::norm_nfd, utf8);
+    // Non-locale aware lowercase
+    boost::algorithm::to_lower(normalized);
+
```
kaykurokawa: looks like you were testing this and forgot to change?

BrannonKing:
If I try to run on mainnet with a fork height that is past I crash with this error:
```
EXCEPTION: 15dbwrapper_error       
Database I/O error       
lbrycrd in AppInit() 
```

src/claimtrie.h
lbrynaut: No, I left it open so it can be decided by all of us.  Locale-aware or not. Choose wisely ;-)

kaykurokawa: According to  http://charette.no-ip.com:81/programming/doxygen/boost/namespaceboost_1_1algorithm.html#a1a318a47f9b4482c28552a62d0297a27 , it seems to be using std::locale() which is whatever the global locale is. 

And it seems like the global locale always defaults to  the "C" locale for C++ programs? (At least that's what it is set to for me, and other sources seem to say that the global locale should always be "C" )

src/claimtrie.h
```diff
@@ -28,6 +34,34 @@
...
+    // Locale aware lowercase
+    /* normalized = boost::locale::to_lower(normalized, utf8); */
+
+    LogPrintf("%s: Normalized \"%s\" (%lu, %lu) to \"%s\" (%lu, %lu)\n", __func__, name, name.size(), sizeof(name), normalized, normalized.size(), sizeof(normalized));
+    // code points can be lost if we're not careful to make sure the string isn't shrinking here
+    return (normalized.size() >= name.size()) ? normalized : name;
```
kaykurokawa: I don't understand exactly what can cause a string to shrink here?

src/claimtrie.h
lbrynaut: Right, C is usually the default, which is not UTF-8.  The goal is to be systematic (i.e. consistently using normalization), and then applying some effort on lower casing (where a perfect system for all encodings doesn't exist).  So long as we're consistent, it's best effort IMO.  I'm open to other takes on it.

src/claimtrie.h
lbrynaut: That's for UTF-8 code points after normalization.  For example, 0xc0a0 and 0x20 are encodings of the same thing, but collapsing the former (due to the lower casing) is not handled as equivalent.  When there's a dispute, we use the original.  Perhaps we could opt to use something else in that case, but again, that's for consistency.

lbrynaut: > I don't understand exactly what can cause a string to shrink here? 

Also, technically it isn't a string as you normally think of them anymore.  It is a string of byte-length code points, which are variably sized to make the characters.  Sometimes a single encoded character may be N bytes, but after normalization it may be M.

kaykurokawa: Let me rephrase, I get why the size of the un-normalized string could shrink after normalization, 

But I don't get why we are using that as the basis for determining whether we return the original un-normalized string, or the normalized string? Is there something special about this rule? Can you elaborate on your specific example of 0x20 and 0xc0a0 here? What are those characters and why aren't they handled as equivalent?

kaykurokawa:
I think we should discuss what happens to bytes that are invalid utf-8 . Because utf-8 is variable width, not all bytes are valid utf-8. It seems like the normalization function we have now will parse invalid bytes without throwing an error. I'm not sure exactly what the logic is but if I try passing in a byte that is invalid like 0xFF to the normalization function, it returns back 0xFF because it hits this below line in the normalization function:

`return (normalized.size() >= name.size()) ? normalized : name;`

The variable "normalized" was actually empty when I inspected it. I create this unit test gist, for testing purposes, you can check it out: https://gist.github.com/kaykurokawa/2c35843eb09da2bc8f31ffac46de4099

So two questions. First question is whether we should attempt to normalize invalid bytes at all, perhaps we should reject them from ever entering the claimtrie? Second question is if we are normalizing invalid bytes, than is this current method correct (is it possible that invalid bytes could be accidentally normalized into valid bytes?)

BrannonKing:
I tried to test this in a simple fashion this morning with no luck:
1. Remove `~/.lbrycrd/regtest`
2. Compile with regtest normalization fork height at 20.
3. Run the daemon: `./src/lbrycrdd -server -txindex -regtest -walletbroadcast=0`
4. Generate 30 blocks: `./lbrycrd/src/lbrycrd-cli -regtest generate 30`
5. Get this error: 
```
error code: -1
error message:
Database I/O error
```

lbrynaut:
> So two questions. First question is whether we should attempt to normalize invalid bytes at all, perhaps we should reject them from ever entering the claimtrie? Second question is if we are normalizing invalid bytes, than is this current method correct (is it possible that invalid bytes could be accidentally normalized into valid bytes?)

@kaykurokawa Good questions.  There are few issues that need addressing:

1) Normalization can fail
2) Lower-casing can fail
[ Generically, we call both steps normalization in the code ]

The general approach I was going for is that if either fails, fall-back to our current behavior (which is to just treat the bytes as bytes in the claimtrie).  Maybe that was a bad assumption on what we need, or maybe it's not working as intended (although it appears to be to me).

Glad we're finally doing this kind of review though because I did expect this process to take some time and we're getting to the hard parts.  Open to suggestions.  I'm not sure rejecting any non-UTF8 inputs is the correct solution, as it does place a larger burden on the applications using lbrycrd, who really may not care about this feature.

Either way, I think we can all agree that consistency is the most important, in that either the same byte or code-point sequences are stored consistently in the claimtrie.

src/main.cpp
```diff
@@ -2735,13 +2762,23 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         pos.nTxOffset += ::GetSerializeSize(tx, SER_DISK, CLIENT_VERSION);
     }
 
-    assert(trieCache.incrementBlock(blockundo.insertUndo, blockundo.expireUndo, blockundo.insertSupportUndo, blockundo.expireSupportUndo, blockundo.takeoverHeightUndo));
+    if (pindex->nHeight == Params().GetConsensus().nNormalizedNameForkHeight) {
```
BrannonKing: Is there an advantage to having this code to enable the normalization go here rather than in the CClaimTrieCache object (since the latter also gets IncrementBlock called)? What rule would guide that?

src/main.cpp
```diff
@@ -2735,13 +2762,23 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+        LogPrintf("Incremented past the normalization hard fork height\n");
+        pclaimTrie = pclaimTrie->enableNormalizationFork(true, pindex->nHeight, trieCache);
+        CClaimTrieCache tmpTrieCache(pclaimTrie);
+        tmpTrieCache.setBestBlock(trieCache.getBestBlock());
+        trieCache = tmpTrieCache;
```
BrannonKing: Why do we need to make a new trieCache object here?

src/claimtrie.cpp
```diff
@@ -137,6 +156,109 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
...
+        LogPrintf("Pre-normalization fork claim trie nodes = %lu\n", nodes.size());
+
+        if (normalizationForkStateInstance.nonNormalizedTrie == NULL) {
+            normalizationForkStateInstance.nonNormalizedTrie = this;
+        }
+
+        if (normalizationForkStateInstance.nonNormalizedTrieCache == NULL) {
+            normalizationForkStateInstance.nonNormalizedTrieCache = boost::make_shared<CClaimTrieCache>(this);
+            normalizationForkStateInstance.nonNormalizedTrieCache->cloneState(trieCache);
+            LogPrintf("%s: Cloned Merkle hash %s, Original %s\n", __func__, normalizationForkStateInstance.nonNormalizedTrieCache->getMerkleHash(true).GetHex().c_str(), trieCache.getMerkleHash(true).GetHex().c_str());
```
BrannonKing: This line is brutally expensive.

src/claimtrie.cpp
```diff
@@ -137,6 +156,109 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
...
+            LogPrintf("Post-normalization fork claim trie nodes = %lu\n", nodes.size());
+        }
+
+        assert(normalizationForkStateInstance.nonNormalizedTrie);
+        return normalizationForkStateInstance.normalizedTrie;
+    }
+
+    if (normalizationForkStateInstance.nonNormalizedTrieCache != NULL) {
+        assert(normalizationForkStateInstance.nonNormalizedTrieCache != NULL);
+        LogPrintf("%s: Restoring cloned Merkle hash %s, Empty %s\n", __func__, normalizationForkStateInstance.nonNormalizedTrieCache->getMerkleHash(true).GetHex().c_str(), trieCache.getMerkleHash(true).GetHex().c_str());
```
BrannonKing: Again, regenerating the hashes for the entire trie (x2) is too expensive for logging.

src/claimtrie.cpp
BrannonKing: What if the existing node had additional claims on it?

src/claimtrie.cpp
```diff
@@ -1112,20 +1263,23 @@ bool CClaimTrie::WriteToDisk()
...
 {
-    if (name.size() == 0)
-    {
+    if (name.size() == 0) {
         root = *node;
         return true;
     }
+    const std::string& newName = (fNormalize ? normalizeClaimName(name) : name);
+    if (fNormalize && newName != name) {
+        delete node;
```
BrannonKing: This line looks suspicious to me; what if the previous node had claims on it? It appears that we don't keep those. Is this where we're losing claims? We need to change them so that all the original ones still complete; don't throw them away.

src/claimtrie.cpp
```diff
@@ -1112,20 +1263,23 @@ bool CClaimTrie::WriteToDisk()
...
+        node = new CClaimTrieNode();
+    }
     CClaimTrieNode* current = &root;
-    for (std::string::const_iterator itname = name.begin(); itname + 1 != name.end(); ++itname)
-    {
+    for (std::string::const_iterator itname = newName.begin(); itname + 1 != newName.end(); ++itname) {
         nodeMapType::iterator itchild = current->children.find(*itname);
         if (itchild == current->children.end())
             return false;
```
BrannonKing: I know this is a preexisting line, but it seems to make a harsh assumption about the order that data in the trie is read from disk (it must be "pre-order").

src/claimtrie.cpp
```diff
@@ -137,6 +156,109 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
...
+            normalizationForkStateInstance.nonNormalizedTrieCache = boost::make_shared<CClaimTrieCache>(this);
+            normalizationForkStateInstance.nonNormalizedTrieCache->cloneState(trieCache);
+            LogPrintf("%s: Cloned Merkle hash %s, Original %s\n", __func__, normalizationForkStateInstance.nonNormalizedTrieCache->getMerkleHash(true).GetHex().c_str(), trieCache.getMerkleHash(true).GetHex().c_str());
+        }
+
+        if (normalizationForkStateInstance.normalizedTrie == NULL) {
+            normalizationForkStateInstance.normalizedTrie = new CClaimTrie(true, false, true);
+            assert(normalizationForkStateInstance.normalizedTrie->shouldNormalize());
+            normalizationForkStateInstance.nonNormalizedTrie->WriteToDisk();
+            normalizationForkStateInstance.normalizedTrie->ReadFromDisk(true);
```
BrannonKing: Thinking about this some: an alternative to this would be to simply walk the trie and add necessary changes to the insertion and deletion queues. As written this seems to make an assumption that the vast majority of the trie will have to change. If we modify the cache instead the persisted undo info would allow us to rollback without having to store the duplicated trie in RAM.

src/claimtrie.cpp
```diff
@@ -137,6 +156,109 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
...
+    for (expirationQueueType::const_iterator it = other.dirtyExpirationQueueRows.begin(); it != other.dirtyExpirationQueueRows.end(); ++it)
+        dirtyExpirationQueueRows[it->first] = it->second;
+    for (supportQueueType::const_iterator it = other.dirtySupportQueueRows.begin(); it != other.dirtySupportQueueRows.end(); ++it)
+        dirtySupportQueueRows[it->first] = it->second;
+    for (expirationQueueType::const_iterator it = other.dirtySupportExpirationQueueRows.begin(); it != other.dirtySupportExpirationQueueRows.end(); ++it)
+        dirtySupportExpirationQueueRows[it->first] = it->second;
+    for (nodeCacheType::const_iterator it = other.dirtyNodes.begin(); it != other.dirtyNodes.end(); ++it)
+        dirtyNodes[normalizeClaimName(it->first)] = new CClaimTrieNode();
+    for (supportMapType::const_iterator it = other.dirtySupportNodes.begin(); it != other.dirtySupportNodes.end(); ++it)
+        dirtySupportNodes[it->first] = it->second;
```
BrannonKing: This line is missing the `normalizedClaimName` on `it->first` ?

src/claimtrie.cpp
lbrynaut: In what way?  Show me the measurements.

BrannonKing: We presently have 5.5million nodes in the trie. `getMerkleHash(true)` triggers a rehash on every one of those. I just measured the call at 3 microseconds each. This was on a non-optimized build, but even if you figure 1 microsecond per hash call you're looking at 10 seconds of CPU hit for that one line.

configure.ac
```diff
@@ -746,8 +753,19 @@ if test x$use_pkgconfig = xyes; then
...
+      ICU_CPPFLAGS="-I$ICU_PREFIX/include"
+      CPPFLAGS="$ICU_CPPFLAGS $CPPFLAGS"
+      ICU_LIBS="-L$ICU_PREFIX/lib -licui18n -licuuc -licudata -dl"
+      AC_MSG_WARN([Using ICU_CPPFLAGS $ICU_CPPFLAGS])
+      AC_MSG_WARN([Using CPPFLAGS $CPPFLAGS])
+
+      AC_CHECK_HEADER([unicode/errorcode.h],,AC_MSG_ERROR(libicu headers missing))
+      AC_CHECK_LIB([icudata], [main], ICU_LIBS="-L$ICU_PREFIX/lib -licui18n -licuuc -licudata -dl",
+        AC_CHECK_LIB([icu18n], [main],ICU_LIBS=$ICU_LIBS, ICU_LIBS="-L$ICU_PREFIX/lib -lsicudt -lsicuin -lsicuio -lsicule -lsiculx -lsicutest -lsicutu -lsicuuc"))
+
```
bvbfan: I investigate why configure without --with-icu fails, this section is for pkg-config you can replace all lines 756-765 with
```
PKG_CHECK_MODULES([ICUI18N], [icu-i18n],, [AC_MSG_ERROR(icu-i18n  not found.)])
```
Then it will make all needed flags and libs for compilation, `else` section (792-802) looks correct.

BrannonKing: I just tested this and still got the same error as before: `cannot determine absolute directory name of 'auto/lib'`. I modified configure.ac as suggested above, then reran autogen.sh and configure (`CXXFLAGS="-g" ./configure --enable-cxx --enable-static --disable-shared --with-pic --with-gui=no`) and make.

bvbfan: That's the diff, it works here.
```
diff --git a/configure.ac b/configure.ac
index 3d9f513d9..243edf5b5 100644
--- a/configure.ac
+++ b/configure.ac
@@ -753,15 +753,9 @@ if test x$use_pkgconfig = xyes; then
   m4_ifdef(
     [PKG_CHECK_MODULES],
     [
-      ICU_CPPFLAGS="-I$ICU_PREFIX/include"
-      CPPFLAGS="$ICU_CPPFLAGS $CPPFLAGS"
-      ICU_LIBS="-L$ICU_PREFIX/lib -licui18n -licuuc -licudata -dl"
-      AC_MSG_WARN([Using ICU_CPPFLAGS $ICU_CPPFLAGS])
-      AC_MSG_WARN([Using CPPFLAGS $CPPFLAGS])
+      PKG_CHECK_MODULES([ICUI18N], [icu-i18n],, [AC_MSG_ERROR(icu-i18n  not found.)])
 
       AC_CHECK_HEADER([unicode/errorcode.h],,AC_MSG_ERROR(libicu headers missing))
-      AC_CHECK_LIB([icudata], [main], ICU_LIBS="-L$ICU_PREFIX/lib -licui18n -licuuc -licudata -dl",
-        AC_CHECK_LIB([icu18n], [main],ICU_LIBS=$ICU_LIBS, ICU_LIBS="-L$ICU_PREFIX/lib -lsicudt -lsicuin -lsicuio -lsicule -lsiculx -lsicutest -lsicutu -lsicuuc"))
 
       PKG_CHECK_MODULES([SSL], [libssl],, [AC_MSG_ERROR(openssl  not found.)])
       PKG_CHECK_MODULES([CRYPTO], [libcrypto],,[AC_MSG_ERROR(libcrypto  not found.)])
```

BrannonKing:
I'm getting close to having a solution that doesn't require keeping two tries in memory. As part of that work, I've been narrowing in on two rules:
1. You cannot normalize a string inside of a method that takes a partial key; you run the risk of losing data.
2. You cannot normalize inside of a claimtrie's method called from claimtriecache; they aren't on guaranteed to be on the same block height. (It's possible that we could make this work by careful locations for turning the normalization handling in the claimtrie on/off, but I haven't seen a need for it yet.)

src/claimtrie.cpp
```diff
@@ -2336,13 +2177,8 @@ bool CClaimTrieCache::incrementBlock(insertUndoType& insertUndo, claimQueueRowTy
...
             CClaimValue claim;
-            const std::string& name = (base->shouldNormalize() ? normalizeClaimName(itEntry->name) : itEntry->name);
-            // Normalization may have pruned this claim already, so
-            // expiring it is now impossible.  Assert that
-            // normalization is enabled and the claim no longer exists
-            // in the claimtrie OR that the removal is successful.
-            assert((base->shouldNormalize() && !base->haveClaim(name, itEntry->outPoint)) ||
-                   removeClaimFromTrie(name, itEntry->outPoint, claim, true));
+            const std::string name = base->normalizeClaimName(itEntry->name);
+            assert(removeClaimFromTrie(name, itEntry->outPoint, claim, true));
```
lbrynaut: Removal of this assert is fine (it's no longer needed using this approach).

src/claimtrie.cpp
```diff
@@ -2271,30 +2072,70 @@ expirationQueueType::iterator CClaimTrieCache::getSupportExpirationQueueCacheRow
...
+
+                removeUndo.push_back(std::make_pair(it->first, *itClaim));
+                bool success = removeClaimFromTrie(it->first, itClaim->outPoint, *itClaim, false);
+                assert(success);
+
+                insertUndo.push_back(nameOutPointHeightType(normalized, itClaim->outPoint, itClaim->nValidAtHeight));
+                success = insertClaimIntoTrie(normalized, *itClaim, true);
+                assert(success);
+
+//                claimQueueEntryType entry(normalized, *itClaim);
```
bvbfan: Why commented code?

src/qt/paymentrequestplus.cpp
```diff
@@ -148,38 +142,49 @@ bool PaymentRequestPlus::getMerchant(X509_STORE* certStore, QString& merchant) c
...
         rcopy.SerializeToString(&data_to_verify);
 
-        EVP_MD_CTX ctx;
-        EVP_PKEY *pubkey = X509_get_pubkey(signing_cert);
-        EVP_MD_CTX_init(&ctx);
-        if (!EVP_VerifyInit_ex(&ctx, digestAlgorithm, NULL) ||
-            !EVP_VerifyUpdate(&ctx, data_to_verify.data(), data_to_verify.size()) ||
-            !EVP_VerifyFinal(&ctx, (const unsigned char*)paymentRequest.signature().data(), (unsigned int)paymentRequest.signature().size(), pubkey)) {
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+        EVP_MD_CTX* ctx = EVP_MD_CTX_new();
```
bvbfan: For OpenSSL 1.0 you can use EVP_MD_CTX_create or
```
#if OPENSSL_VERSION_NUMBER >= 0x10000000L && OPENSSL_VERSION_NUMBER < 0x10100000L
#define EVP_MD_CTX_new EVP_MD_CTX_create
#define EVP_MD_CTX_free EVP_MD_CTX_destroy
#endif
Then you can use EVP_MD_CTX_new and EVP_MD_CTX_free unconditional.

File: src/claimtrie.cpp [hunk 37]
```diff
@@ -2491,12 +2346,12 @@ bool CClaimTrieCache::incrementBlock(insertUndoType& insertUndo, claimQueueRowTy
                     {
                         LogPrintf("%s(): An inconsistency was found in the claim queue. Please report this to the developers:\nClaim found in name queue but not in height based queue:\nname: %s, txid: %s, nOut: %d, nValidAtHeight in name based queue: %d, current height: %d\n", __func__, nameToCheck, itQueueName->outPoint.hash.GetHex(), itQueueName->outPoint.n, itQueueName->nHeight, nCurrentHeight);
                     }
-                    assert(found);
+                    //assert(found);
```
bvbfan: If this assert fails it looks like it compares noramalized and not normalized name, that's pretty likely when it used in increment block.

src/claimtrie.cpp
BrannonKing: it was for a previous discussion and now removed

src/qt/paymentrequestplus.cpp
BrannonKing: wasn't meant to be part of this review; make sure #221 got it correct

bvbfan:
normalizeClaimName should be static, inilining here is unwanted, furthermore checking, to be performed normalize, should not be there. For now it should have ability to be called from CClaimTrie and CClaimTrieCache depending of their heights. I made a unit test to verify that
```
BOOST_AUTO_TEST_CASE(claimtriecache_normalization)
{

    // Disable reg-test expiration during this test
    pclaimTrie->setExpirationTime(5000);

    // Add another set of unicode claims that will collapse after the fork
    CMutableTransaction tx = fixture.MakeClaim(fixture.GetCoinbase(), "Ame\u0301lie", "amelie", 2);
    fixture.MakeClaim(fixture.GetCoinbase(), "amelie1", "amelie", 2);
    fixture.IncrementBlocks(1);
    CClaimValue nval1;
    BOOST_CHECK(pclaimTrie->getInfoForName("amelie1", nval1));
    BOOST_CHECK(!pclaimTrie->shouldNormalize());
    // amélie is not found cause normalization still not appear
    BOOST_CHECK(!pclaimTrie->getInfoForName("amélie", nval1));

    // Activate the fork (which rebuilds the existing claimtrie and
    // cache), flattening all previously existing name clashes due to
    // the normalization
    int currentHeight = chainActive.Height();
    const Consensus::Params& consensus = Params().GetConsensus();
    int forkOrig = consensus.nNormalizedNameForkHeight;
    const_cast<Consensus::Params&>(consensus).nNormalizedNameForkHeight = currentHeight + 1;
    BOOST_SCOPE_EXIT(&consensus, forkOrig) {
        const_cast<Consensus::Params&>(consensus).nNormalizedNameForkHeight = forkOrig;
    } BOOST_SCOPE_EXIT_END

    fixture.IncrementBlocks(1, true);
    BOOST_CHECK(pclaimTrie->shouldNormalize());
    // Ok normalization fix the name problem
    BOOST_CHECK(pclaimTrie->getInfoForName("amélie", nval1));
    BOOST_CHECK(nval1.nHeight == currentHeight);

    CBlockUndo blockUndo;
    int amelieValidHeight;
    CClaimTrieCache trieCache(pclaimTrie);
    BOOST_CHECK(trieCache.decrementBlock(blockUndo.insertUndo, blockUndo.expireUndo, blockUndo.insertSupportUndo, blockUndo.expireSupportUndo, blockUndo.takeoverHeightUndo));
    // Even we return before normalization, claim is reachable, i.e. next line fails but it shouldn't
    BOOST_CHECK(!trieCache.spendClaim("amélie", COutPoint(tx.GetHash(), 0), currentHeight, amelieValidHeight));
}
```
Ideally CClaimTrie should not perform any kind of name normalization, it should take care to provide iterable access to trie, depending on its height CClaimTrieCache (it will be renamed, now i face to be CClaimTrieConsumer) shall normalize and add/spend but that's discussion to #106.

It still has compilation issue when --with-icu is not provide, a patch for that
```
diff --git a/src/Makefile.bench.include b/src/Makefile.bench.include
index 538254709..8e7b59b46 100644
--- a/src/Makefile.bench.include
+++ b/src/Makefile.bench.include
@@ -31,7 +31,7 @@ bench_bench_bitcoin_LDADD += $(LIBBITCOIN_WALLET)
 endif
 
 bench_bench_bitcoin_LDADD += $(BOOST_LIBS) $(BDB_LIBS) $(SSL_LIBS) $(CRYPTO_LIBS) $(MINIUPNPC_LIBS) $(EVENT_PTHREADS_LIBS) $(EVENT_LIBS)
-bench_bench_bitcoin_LDFLAGS = -L$(ICU_PREFIX)/lib $(RELDFLAGS) $(AM_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
+bench_bench_bitcoin_LDFLAGS = $(RELDFLAGS) $(AM_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
 
 CLEAN_BITCOIN_BENCH = bench/*.gcda bench/*.gcno
 
diff --git a/src/Makefile.qt.include b/src/Makefile.qt.include
index 17e4d8279..3c742c531 100644
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -382,7 +382,7 @@ endif
 qt_lbrycrd_qt_LDADD += $(LIBBITCOIN_CLI) $(LIBBITCOIN_COMMON) $(LIBBITCOIN_UTIL) $(LIBBITCOIN_CONSENSUS) $(LIBBITCOIN_CRYPTO) $(LIBUNIVALUE) $(LIBLEVELDB) $(LIBMEMENV) \
   $(BOOST_LIBS) $(QT_LIBS) $(QT_DBUS_LIBS) $(QR_LIBS) $(PROTOBUF_LIBS) $(ICU_LIBS) $(BDB_LIBS) $(SSL_LIBS) $(CRYPTO_LIBS) $(MINIUPNPC_LIBS) $(LIBSECP256K1) \
   $(EVENT_PTHREADS_LIBS) $(EVENT_LIBS)
-qt_lbrycrd_qt_LDFLAGS = -L$(ICU_PREFIX)/lib $(RELDFLAGS) $(AM_LDFLAGS) $(QT_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
+qt_lbrycrd_qt_LDFLAGS = $(RELDFLAGS) $(AM_LDFLAGS) $(QT_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
 qt_lbrycrd_qt_LIBTOOLFLAGS = --tag CXX
 
 #locale/foo.ts -> locale/foo.qm
```

bvbfan:
normalizeClaimName should be static, inilining here is unwanted, furthermore checking, to be performed normalize, should not be there. For now it should have ability to be called from CClaimTrie and CClaimTrieCache depending of their heights. I made a unit test to verify that
```
BOOST_AUTO_TEST_CASE(claimtriecache_normalization)
{

    // Disable reg-test expiration during this test
    pclaimTrie->setExpirationTime(5000);

    // Add another set of unicode claims that will collapse after the fork
    CMutableTransaction tx = fixture.MakeClaim(fixture.GetCoinbase(), "Ame\u0301lie", "amelie", 2);
    fixture.MakeClaim(fixture.GetCoinbase(), "amelie1", "amelie", 2);
    fixture.IncrementBlocks(1);
    CClaimValue nval1;
    BOOST_CHECK(pclaimTrie->getInfoForName("amelie1", nval1));
    BOOST_CHECK(!pclaimTrie->shouldNormalize());
    // amélie is not found cause normalization still not appear
    BOOST_CHECK(!pclaimTrie->getInfoForName("amélie", nval1));

    // Activate the fork (which rebuilds the existing claimtrie and
    // cache), flattening all previously existing name clashes due to
    // the normalization
    int currentHeight = chainActive.Height();
    const Consensus::Params& consensus = Params().GetConsensus();
    int forkOrig = consensus.nNormalizedNameForkHeight;
    const_cast<Consensus::Params&>(consensus).nNormalizedNameForkHeight = currentHeight + 1;
    BOOST_SCOPE_EXIT(&consensus, forkOrig) {
        const_cast<Consensus::Params&>(consensus).nNormalizedNameForkHeight = forkOrig;
    } BOOST_SCOPE_EXIT_END

    fixture.IncrementBlocks(1, true);
    BOOST_CHECK(pclaimTrie->shouldNormalize());
    // Ok normalization fix the name problem
    BOOST_CHECK(pclaimTrie->getInfoForName("amélie", nval1));
    BOOST_CHECK(nval1.nHeight == currentHeight);

    CBlockUndo blockUndo;
    int amelieValidHeight;
    CClaimTrieCache trieCache(pclaimTrie);
    BOOST_CHECK(trieCache.decrementBlock(blockUndo.insertUndo, blockUndo.expireUndo, blockUndo.insertSupportUndo, blockUndo.expireSupportUndo, blockUndo.takeoverHeightUndo));
    // Even we return before normalization, claim is reachable, i.e. next line fails but it shouldn't
    BOOST_CHECK(!trieCache.spendClaim("amélie", COutPoint(tx.GetHash(), 0), currentHeight, amelieValidHeight));
}
```
Ideally CClaimTrie should not perform any kind of name normalization, it should take care to provide iterable access to trie, depending on its height CClaimTrieCache (it will be renamed, now i face to be CClaimTrieConsumer) shall normalize and add/spend but that's discussion to #106.

It still has compilation issue when --with-icu is not provide, a patch for that
```
diff --git a/src/Makefile.bench.include b/src/Makefile.bench.include
index 538254709..8e7b59b46 100644
--- a/src/Makefile.bench.include
+++ b/src/Makefile.bench.include
@@ -31,7 +31,7 @@ bench_bench_bitcoin_LDADD += $(LIBBITCOIN_WALLET)
 endif
 
 bench_bench_bitcoin_LDADD += $(BOOST_LIBS) $(BDB_LIBS) $(SSL_LIBS) $(CRYPTO_LIBS) $(MINIUPNPC_LIBS) $(EVENT_PTHREADS_LIBS) $(EVENT_LIBS)
-bench_bench_bitcoin_LDFLAGS = -L$(ICU_PREFIX)/lib $(RELDFLAGS) $(AM_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
+bench_bench_bitcoin_LDFLAGS = $(RELDFLAGS) $(AM_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
 
 CLEAN_BITCOIN_BENCH = bench/*.gcda bench/*.gcno
 
diff --git a/src/Makefile.qt.include b/src/Makefile.qt.include
index 17e4d8279..3c742c531 100644
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -382,7 +382,7 @@ endif
 qt_lbrycrd_qt_LDADD += $(LIBBITCOIN_CLI) $(LIBBITCOIN_COMMON) $(LIBBITCOIN_UTIL) $(LIBBITCOIN_CONSENSUS) $(LIBBITCOIN_CRYPTO) $(LIBUNIVALUE) $(LIBLEVELDB) $(LIBMEMENV) \
   $(BOOST_LIBS) $(QT_LIBS) $(QT_DBUS_LIBS) $(QR_LIBS) $(PROTOBUF_LIBS) $(ICU_LIBS) $(BDB_LIBS) $(SSL_LIBS) $(CRYPTO_LIBS) $(MINIUPNPC_LIBS) $(LIBSECP256K1) \
   $(EVENT_PTHREADS_LIBS) $(EVENT_LIBS)
-qt_lbrycrd_qt_LDFLAGS = -L$(ICU_PREFIX)/lib $(RELDFLAGS) $(AM_LDFLAGS) $(QT_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
+qt_lbrycrd_qt_LDFLAGS = $(RELDFLAGS) $(AM_LDFLAGS) $(QT_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
 qt_lbrycrd_qt_LIBTOOLFLAGS = --tag CXX
 
 #locale/foo.ts -> locale/foo.qm
```

bvbfan:
normalizeClaimName should be static, inilining here is unwanted, furthermore checking, to be performed normalize, should not be there. For now it should have ability to be called from CClaimTrie and CClaimTrieCache depending of their heights. I made a unit test to verify that
```
BOOST_AUTO_TEST_CASE(claimtriecache_normalization)
{

    // Disable reg-test expiration during this test
    pclaimTrie->setExpirationTime(5000);

    // Add another set of unicode claims that will collapse after the fork
    CMutableTransaction tx = fixture.MakeClaim(fixture.GetCoinbase(), "Ame\u0301lie", "amelie", 2);
    fixture.MakeClaim(fixture.GetCoinbase(), "amelie1", "amelie", 2);
    fixture.IncrementBlocks(1);
    CClaimValue nval1;
    BOOST_CHECK(pclaimTrie->getInfoForName("amelie1", nval1));
    BOOST_CHECK(!pclaimTrie->shouldNormalize());
    // amélie is not found cause normalization still not appear
    BOOST_CHECK(!pclaimTrie->getInfoForName("amélie", nval1));

    // Activate the fork (which rebuilds the existing claimtrie and
    // cache), flattening all previously existing name clashes due to
    // the normalization
    int currentHeight = chainActive.Height();
    const Consensus::Params& consensus = Params().GetConsensus();
    int forkOrig = consensus.nNormalizedNameForkHeight;
    const_cast<Consensus::Params&>(consensus).nNormalizedNameForkHeight = currentHeight + 1;
    BOOST_SCOPE_EXIT(&consensus, forkOrig) {
        const_cast<Consensus::Params&>(consensus).nNormalizedNameForkHeight = forkOrig;
    } BOOST_SCOPE_EXIT_END

    fixture.IncrementBlocks(1, true);
    BOOST_CHECK(pclaimTrie->shouldNormalize());
    // Ok normalization fix the name problem
    BOOST_CHECK(pclaimTrie->getInfoForName("amélie", nval1));
    BOOST_CHECK(nval1.nHeight == currentHeight);

    CBlockUndo blockUndo;
    int amelieValidHeight;
    CClaimTrieCache trieCache(pclaimTrie);
    BOOST_CHECK(trieCache.decrementBlock(blockUndo.insertUndo, blockUndo.expireUndo, blockUndo.insertSupportUndo, blockUndo.expireSupportUndo, blockUndo.takeoverHeightUndo));
    // Even we return before normalization, claim is reachable, i.e. next line fails but it shouldn't
    BOOST_CHECK(!trieCache.spendClaim("amélie", COutPoint(tx.GetHash(), 0), currentHeight, amelieValidHeight));
}
```
Ideally CClaimTrie should not perform any kind of name normalization, it should take care to provide iterable access to trie, depending on its height CClaimTrieCache (it will be renamed, now i face to be CClaimTrieConsumer) shall normalize and add/spend but that's discussion to #106.

It still has compilation issue when --with-icu is not provide, a patch for that
```
diff --git a/src/Makefile.bench.include b/src/Makefile.bench.include
index 538254709..8e7b59b46 100644
--- a/src/Makefile.bench.include
+++ b/src/Makefile.bench.include
@@ -31,7 +31,7 @@ bench_bench_bitcoin_LDADD += $(LIBBITCOIN_WALLET)
 endif
 
 bench_bench_bitcoin_LDADD += $(BOOST_LIBS) $(BDB_LIBS) $(SSL_LIBS) $(CRYPTO_LIBS) $(MINIUPNPC_LIBS) $(EVENT_PTHREADS_LIBS) $(EVENT_LIBS)
-bench_bench_bitcoin_LDFLAGS = -L$(ICU_PREFIX)/lib $(RELDFLAGS) $(AM_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
+bench_bench_bitcoin_LDFLAGS = $(RELDFLAGS) $(AM_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
 
 CLEAN_BITCOIN_BENCH = bench/*.gcda bench/*.gcno
 
diff --git a/src/Makefile.qt.include b/src/Makefile.qt.include
index 17e4d8279..3c742c531 100644
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -382,7 +382,7 @@ endif
 qt_lbrycrd_qt_LDADD += $(LIBBITCOIN_CLI) $(LIBBITCOIN_COMMON) $(LIBBITCOIN_UTIL) $(LIBBITCOIN_CONSENSUS) $(LIBBITCOIN_CRYPTO) $(LIBUNIVALUE) $(LIBLEVELDB) $(LIBMEMENV) \
   $(BOOST_LIBS) $(QT_LIBS) $(QT_DBUS_LIBS) $(QR_LIBS) $(PROTOBUF_LIBS) $(ICU_LIBS) $(BDB_LIBS) $(SSL_LIBS) $(CRYPTO_LIBS) $(MINIUPNPC_LIBS) $(LIBSECP256K1) \
   $(EVENT_PTHREADS_LIBS) $(EVENT_LIBS)
-qt_lbrycrd_qt_LDFLAGS = -L$(ICU_PREFIX)/lib $(RELDFLAGS) $(AM_LDFLAGS) $(QT_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
+qt_lbrycrd_qt_LDFLAGS = $(RELDFLAGS) $(AM_LDFLAGS) $(QT_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
 qt_lbrycrd_qt_LIBTOOLFLAGS = --tag CXX
 
 #locale/foo.ts -> locale/foo.qm
```

bvbfan:
normalizeClaimName should be static, inilining here is unwanted, furthermore checking, to be performed normalize, should not be there. For now it should have ability to be called from CClaimTrie and CClaimTrieCache depending of their heights. I made a unit test to verify that
```
BOOST_AUTO_TEST_CASE(claimtriecache_normalization)
{

    // Disable reg-test expiration during this test
    pclaimTrie->setExpirationTime(5000);

    // Add another set of unicode claims that will collapse after the fork
    CMutableTransaction tx = fixture.MakeClaim(fixture.GetCoinbase(), "Ame\u0301lie", "amelie", 2);
    fixture.MakeClaim(fixture.GetCoinbase(), "amelie1", "amelie", 2);
    fixture.IncrementBlocks(1);
    CClaimValue nval1;
    BOOST_CHECK(pclaimTrie->getInfoForName("amelie1", nval1));
    BOOST_CHECK(!pclaimTrie->shouldNormalize());
    // amélie is not found cause normalization still not appear
    BOOST_CHECK(!pclaimTrie->getInfoForName("amélie", nval1));

    // Activate the fork (which rebuilds the existing claimtrie and
    // cache), flattening all previously existing name clashes due to
    // the normalization
    int currentHeight = chainActive.Height();
    const Consensus::Params& consensus = Params().GetConsensus();
    int forkOrig = consensus.nNormalizedNameForkHeight;
    const_cast<Consensus::Params&>(consensus).nNormalizedNameForkHeight = currentHeight + 1;
    BOOST_SCOPE_EXIT(&consensus, forkOrig) {
        const_cast<Consensus::Params&>(consensus).nNormalizedNameForkHeight = forkOrig;
    } BOOST_SCOPE_EXIT_END

    fixture.IncrementBlocks(1, true);
    BOOST_CHECK(pclaimTrie->shouldNormalize());
    // Ok normalization fix the name problem
    BOOST_CHECK(pclaimTrie->getInfoForName("amélie", nval1));
    BOOST_CHECK(nval1.nHeight == currentHeight);

    CBlockUndo blockUndo;
    int amelieValidHeight;
    CClaimTrieCache trieCache(pclaimTrie);
    BOOST_CHECK(trieCache.decrementBlock(blockUndo.insertUndo, blockUndo.expireUndo, blockUndo.insertSupportUndo, blockUndo.expireSupportUndo, blockUndo.takeoverHeightUndo));
    // Even we return before normalization, claim is reachable, i.e. next line fails but it shouldn't
    BOOST_CHECK(!trieCache.spendClaim("amélie", COutPoint(tx.GetHash(), 0), currentHeight, amelieValidHeight));
}
```
Ideally CClaimTrie should not perform any kind of name normalization, it should take care to provide iterable access to trie, depending on its height CClaimTrieCache (it will be renamed, now i face to be CClaimTrieConsumer) shall normalize and add/spend but that's discussion to #106.

BrannonKing:
@bvbfan , excellent analysis. With our work on #44 it appears that we achieved independence from methods taking a name in rpc/claimtrie.cpp. I think that is critical to your suggestion that we don't do string normalization in the CClaimTrie (instead, we do it all in the "cache"). I really like the suggestion! I think that puts a high priority on getting #44 merged and rebasing our normalization on top of that.

src/claimtrie.h
```diff
@@ -391,19 +306,56 @@ class CClaimTrieCache;
...
+        nExpirationTime(Params().GetConsensus().nOriginalClaimExpirationTime),
+        nProportionalDelayFactor(nProportionalDelayFactor),
+        root(uint256S("0000000000000000000000000000000000000000000000000000000000000001"))
     {}
 
-    // Does not change the current trie, but returns a normalized equivalent trie
-    CClaimTrie* enableNormalizationFork(bool enable, int nHeight, CClaimTrieCache& trieCache);
+    // lower-case and normalize any input string name
+    // see: https://unicode.org/reports/tr15/#Norm_Forms
+    inline std::string normalizeClaimName(const std::string& name, bool force = false) const
```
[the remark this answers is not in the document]
BrannonKing: fixed in last push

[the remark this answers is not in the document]
BrannonKing: fixed

src/claimtrie.cpp
BrannonKing: Fixed in current push.

src/claimtrie.cpp
```diff
@@ -1623,14 +1610,14 @@ bool CClaimTrieCache::removeClaimFromQueue(const std::string& name, const COutPo
 bool CClaimTrieCache::undoAddClaim(const std::string& name, const COutPoint& outPoint, int nHeight) const
 {
     int throwaway;
-    const std::string newName = base->normalizeClaimName(name);
+    const std::string newName = normalizeClaimName(name);
     return removeClaim(newName, outPoint, nHeight, throwaway, false);
 }
 
-bool CClaimTrieCache::spendClaim(const std::string& name, const COutPoint& outPoint, int nHeight, int& nValidAtHeight) const
+bool CClaimTrieCache::spendClaim(std::string& name, const COutPoint& outPoint, int nHeight, int& nValidAtHeight) const
```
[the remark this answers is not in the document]
BrannonKing: The out parameter is used in mining.cpp and main.cpp (to avoid duplicate work that name). Out parameter usage is not very obvious, though. I'm open to other suggestions.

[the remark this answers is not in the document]
bvbfan: Yeah, i see now. The right approach, to me, is to call normalize before spend, we can prevent double work if we have function like isNormlized before real normalization. But i don't think double call will slow down significantly.

reproducible_build.sh
```diff
@@ -348,19 +384,27 @@ function build_lbrycrd() {
...
+
+    CPPFLAGS="${CPPFLAGS}" LDFLAGS="${LDFLAGS}"  \
+            ./configure --without-gui ${OPTIONS} \
+	    --with-boost="${BOOST_PREFIX}"   \
+            --with-icu="${ICU_PREFIX}" >> "${LBRYCRD_LOG}" 2>&1
+
+    
+    background "make ${PARALLEL}" "${LBRYCRD_LOG}" "Waiting for lbrycrd to finish building"
+    if [ "${OS_NAME}" = "linux" ]; then
+        LD_LIBRARY_PATH="${ICU_PREFIX}/lib" src/test/test_lbrycrd
```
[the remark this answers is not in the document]
bvbfan: Notice line 333 it has --enable-shared, so building with this option the result is dynamic libraries (*.so) but for static build we want a static ones (*.a) so it should be --disable-shared --enable-static.

src/claimtrie.cpp
```diff
@@ -2819,3 +2859,49 @@ bool CClaimTrieCache::forkForExpirationChange(bool increment) const
     return true;
 }
 
+bool CClaimTrieCache::shouldNormalize() const {
```
bvbfan: Function opening brace goes to the new line, that's tidy expectation.

src/claimtrie.cpp
```diff
@@ -2819,3 +2859,49 @@ bool CClaimTrieCache::forkForExpirationChange(bool increment) const
...
+    return nCurrentHeight > Params().GetConsensus().nNormalizedNameForkHeight;
+}
+
+std::string CClaimTrieCache::normalizeClaimName(const std::string& name, bool force) const {
```
bvbfan: Opening brace on new line.

src/claimtrie.cpp
bvbfan: Yes, it will.

reproducible_build.sh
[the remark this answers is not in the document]
BrannonKing: fixed

BrannonKing:
I did finally get a green on the Linux compilation again. Some notes on my issues with the reproducible_build.sh:
1. The ICU_PREFIX was being exported to the child makefiles but wasn't always set to a real value when relying on pkg-config. This was causing issues. I changed the child makefiles to rely wholly upon ICU_LIBS and ICU_CPPFLAGS, which should be always set.
2. I could not get ICU v55 to compile statically. I upped it to v57. Our current version of Boost (v59) would not compile with newer versions of ICU.
3. The script sets "-e", or maybe it's "-u", which makes it exit whenever any command returns nonzero. It also sets pipeline, which makes it keep the worst result from any method. The flag handling in the wait method was always returning 1 on my machine, causing the whole script to exit. I ditched the flag usage there in favor of a sub-terminal.
4. The parameters for the boost b2 command weren't getting grouped appropriately. I put in quotes. This made it so that I could not make that call from the "background" function. I'm hoping someone else has a solution for that. The background function seems to run it fine, it just doesn't find the ICU deps when called from there. I could not figure that out. I did add a grep check to ensure that the ICU deps are found.
5. Because I'm now compiling Boost without the background minute message, TravisCI times out at 10 minutes when compiling Boost afresh. I don't have a solution for this yet. It wants output every 10 minutes. It's ridiculous that the timeout is not configurable. Instead, we push the output to a log file for each dependency build. I'm not sure that's actually a desirable feature. The other side affect of that is how we hide any error reported by our script with a postdump of 200 or 1000 lines of the log file.
6. The TravisCI caching keeps the build folder around. Our script was checking to see if the dependency's parent folder existed there. If it did exist, it would not run make on that dependency. Hence, if you fail on a dependency it won't get rebuilt because of the TravisCI cache. Not only that, but build/boost was always there -- even if I deleted the cached file from TravisCI. I had to change the script to always build dependencies, which is fast for all except openssl who thinks it's cool to rewrite a whole bunch of source code files with every configure call. 
7. The OSX build now fails with this error: "Too many levels of symbolic links" while compiling openssl a second time. I don't have a solution for that yet.
8. Statically linking in ICU requires fPIC. I had to add that to the build of ICU.
9. The "clone" flag that didn't do anything is now gone.
10. The reproducible_build originally used pkg-config to find ICU, but it would possibly return a system installation. It's now hard-coded to the one pulled by the script.

BrannonKing:
#235 supersedes this PR.

PR #160, opened 2018-06-15
Rename CClaimTrieCache to CClaimTrieUpdateBuffer and separate in own file

bvbfan:


ekxoiner:
How to get start claim lbry coin?

bvbfan:
This a code review can you ask here https://www.facebook.com/lbryio

kaykurokawa:
CClaimTrie.nCurrentHeight has also been renamed to nNextHeight here

src/test/claimtrieupdatebuffer_tests.cpp
```diff
@@ -67,18 +68,18 @@ BOOST_AUTO_TEST_CASE(recursiveprune_test)
     base_node.insertClaim(test_claim);
 
     // node 1 has a claim so it should not be pruned
-    CClaimTrieNode node_1;
+    CClaimTrieNode *node_1 = new CClaimTrieNode;
```
BrannonKing: The old code looked more correct to me. Why the change to an allocation? The modern c++ says that the `new` keyword is generally deprecated. Use `std/boost::make_unique` or `make_shared` if you really need to allocate on the heap, but try to avoid it generally.

bvbfan: Children are vector of shared pointers

bvbfan: new is completely not deprecated or so on.

kaykurokawa:
I had some problem rebasing this, there is basically a commit here:
7d321ec0eac0a9cb14a3ef078f134a87f6aed956 which looks like a failed merge commit that contained elements from unrelated commits. 

I've resolved this in  https://github.com/lbryio/lbrycrd/pull/175 so will be closing this one. 

In the future, I suggest that you rebase commits that you are working on top of master instead of merging it.

Issue #157, opened 2018-06-06
Specify coding style, other guidelines in README


kaykurokawa:
We should probably adopt the same coding style as adopted in Bitcoin Core
https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md

Additional notes: 
Avoid changing components in Bitcoin Core unless it is absolutely necessary (basically if the codes does not have anything to do with the claimtrie, it is part of Bitcoin Core although there are exceptions such as the POW algorithm, reward schedule, various chain params, etc..). This is to make merges from Bitcoin Core as easy as possible.

kaykurokawa:
Useful script: 
run this in the base lbrycrd folder ("HEAD~1 can be replaced to look at whatever commit you want") to perform the clang format diff correction
git diff -U0 HEAD~1.. | ./contrib/devtools/clang-format-diff.py -p1 -i -v

kaykurokawa:
implemented in https://github.com/lbryio/lbrycrd/pull/163 and https://github.com/lbryio/lbrycrd/pull/155

PR #163, opened 2018-06-20
Run code formatting checks on travis

kaykurokawa:
Added an option on reproducible_build.sh to run a code formatting check on any commits not on master (by default this option is off). Travis will use this option and build will fail if it finds any lines that needs to be formatted.

For https://github.com/lbryio/lbrycrd/issues/157

### DIFF
--- a/reproducible_build.sh
+++ b/reproducible_build.sh
@@ -16,6 +16,7 @@ function HELP {
     echo "Optional arguments:"
     echo
     echo "-c: don't clone a fresh copy of the repo"
+    echo "-f: check formatting of committed code relative to master"
     echo "-r: remove intermediate files."
     echo "-l: build only lbrycrd"
     echo "-d: build only the dependencies"
@@ -27,6 +28,7 @@ function HELP {
 
 CLONE=true
 CLEAN=false
+CHECK_CODE_FORMAT=false
 BUILD_DEPENDENCIES=true
 BUILD_LBRYCRD=true
 TIMEOUT=false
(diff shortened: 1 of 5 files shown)


lyoshenka:
:+1: this is the only way to enforce code styles, imho

kaykurokawa:
Brought in the latest .clang-format and developer docs from Bitcoin Core. We should follow these from now on.

reproducible_build.sh has a new option flag -f that will run a code formatting check on any commit not on master using .clang-format before building. On travis, the build will fail if there are any code formatting issues on your commits.

If you want to run just the code formatting check script, I added a line in README on how to do that. 

Added some comments on README about dealing with changes that we bring in from Bitcoin Core that is not expected to be used when we do a full merge from upstream (i.e. updating to the lastet Core version). Note that e9369eb is one of those commits.

reproducible_build.sh
```diff
@@ -357,6 +368,13 @@ function build_lbrycrd() {
     strip src/test/test_lbrycrd
 }
 
+function clang_format_diff(){
+    # run a code formatting check on any commits not in master
+    # requires clang-format
+    git fetch origin master
+    git diff -U0 origin/master -- '*.h' '*.cpp' | ./contrib/devtools/clang-format-diff.py -p1
```
lbrynaut: I don't think this will do what you expect it to do.

Try:

```
git remote add origin2 https://github.com/lbryio/lbrycrd.git
git fetch origin2
git diff -U0 origin2/master -- ... snip ...
```

lbrynaut:
@kaykurokawa This looks good to me (although what happens if the style fails -- just a travis failure?).  If you want to merge and create the release including the OS X build, maybe this is a good spot to do it.

Issue #125, opened 2018-04-23
Sanitize RPC commands


shyba:
It's more of a low priority feature request and maybe a good first issue, but something I noticed is that we could validate the parameters in a better way.

Some of our methods doesn't check for validity of claim id or how the parameters comes: https://github.com/lbryio/lbrycrd/blob/master/src/rpc/claimtrie.cpp#L498

In opposition, some of Bitcoin commands does the check and returns an error if it's not a proper hash. For instance `getrawtransaction` calls `ParseHashV` for checking it: https://github.com/lbryio/lbrycrd/blob/270307a29070a4c10095528e232ad0f8e251ec5a/src/rpc/rawtransaction.cpp#L183

Which is implemented here, where we could also add our validators: https://github.com/lbryio/lbrycrd/blob/270307a29070a4c10095528e232ad0f8e251ec5a/src/rpc/server.cpp#L144

Aside from the usability of the cli for people giving the wrong input to commands, this should avoid calling methods over bogus data with sometimes unpredictable outcomes.

kauffj:
Thank you @shyba! And yes, this is a good first issue.

If it's not too much work, could you add a list of methods that need this, or a way someone else could identify the ones that do?

shyba:
Sure!
Commands are implemented here: https://github.com/lbryio/lbrycrd/blob/master/src/rpc/claimtrie.cpp
The ones that needs sanitizing are:
 - [ ] - getnameproof (blockhash)
 - [ ] - getclaimsfortx (txid)
 - [ ] - getclaimbyid (claim id)

See how it's done in here: https://github.com/lbryio/lbrycrd/blob/270307a29070a4c10095528e232ad0f8e251ec5a/src/rpc/rawtransaction.cpp#L183

For the name parameter in `getclaimsforname` and `getvalueforname` no sanitization is necessary since they are expected to be strings in any format.

BrannonKing:
Fixed via https://github.com/lbryio/lbrycrd/commit/e3813362815556634046970d978ba66cef84fea9

PR #167, opened 2018-06-30
Sanitize claimtrie RPC commands.

winder:
Change for issue #125

I wasn't exactly sure how you deal with upstream test changes, but I'm pretty sure this isn't it. For now, I went ahead and updated the existing `rpc_tests` suite instead of copying over the CallRPC utility. Is there a better spot, or approach?

### DIFF
--- a/src/rpc/claimtrie.cpp
+++ b/src/rpc/claimtrie.cpp
@@ -380,6 +380,10 @@ UniValue getclaimbyid(const UniValue& params, bool fHelp)
         );
 
     LOCK(cs_main);
+
+    // Used for validation.
+    ParseHashV(params[0], "Claim-id (parameter 1)");
+
     uint160 claimId;
     claimId.SetHex(params[0].get_str());
     UniValue claim(UniValue::VOBJ);
@@ -512,10 +516,7 @@ UniValue getclaimsfortx(const UniValue& params, bool fHelp)
         );
 
     LOCK(cs_main);
-
-    uint256 hash;
-    hash.SetHex(params[0].get_str());
-
+    uint256 hash = ParseHashV(params[0], "txid (parameter 1)");
     UniValue ret(UniValue::VARR);
 
     int op;
@@ -741,8 +742,7 @@ UniValue getnameproof(const UniValue& params, bool fHelp)
     uint256 blockHash;
     if (params.size() == 2)
     {
-        std::string strBlockHash = params[1].get_str();
-        blockHash = uint256S(strBlockHash);
+        uint256 hash = ParseHashV(params[1], "blockhash (optional parameter 2)");
     }
     else
     {
--- a/src/rpc/claimtrie.cpp
+++ b/src/rpc/claimtrie.cpp
@@ -382,7 +382,7 @@ UniValue getclaimbyid(const UniValue& params, bool fHelp)
     LOCK(cs_main);
 
     // Used for validation.
-    ParseHashV(params[0], "Claim-id (parameter 1)");
+    ParseHexV(params[0], "Claim-id (parameter 1)");
 
     uint160 claimId;
     claimId.SetHex(params[0].get_str());
@@ -742,7 +742,7 @@ UniValue getnameproof(const UniValue& params, bool fHelp)
     uint256 blockHash;
     if (params.size() == 2)
     {
-        uint256 hash = ParseHashV(params[1], "blockhash (optional parameter 2)");
+        blockHash = ParseHashV(params[1], "blockhash (optional parameter 2)");
     }
     else
     {
--- a/src/rpc/claimtrie.cpp
+++ b/src/rpc/claimtrie.cpp
@@ -7,6 +7,22 @@
 // Maximum block decrement that is allowed from rpc calls
 const int MAX_RPC_BLOCK_DECREMENTS = 50;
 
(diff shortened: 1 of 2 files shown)


tzarebczan:
@winder , thanks for this contribution! Please review our [contributing guide](https://lbry.io/faq/contributing) if you have not done so already. Also, [click here](https://lbry.io/faq/tips) if you are interested in an LBC tip for your work :)

We'll also be adding you to our developer mailing list so we can keep in touch (you can unsubscribe anytime!). 

We'll get a proper review on this PR from one of the devs.

src/rpc/claimtrie.cpp
```diff
@@ -380,6 +380,10 @@ UniValue getclaimbyid(const UniValue& params, bool fHelp)
         );
 
     LOCK(cs_main);
+
+    // Used for validation.
+    ParseHashV(params[0], "Claim-id (parameter 1)");
```
winder: This check is probably wrong. Are IDs a hex value or numeric?

lbrynaut: https://github.com/lbryio/lbrycrd/blob/master/src/rpc/claimtrie.cpp#L359

ClaimIds are the hashes represented as hex strings.

src/rpc/claimtrie.cpp
```diff
@@ -741,8 +742,7 @@ UniValue getnameproof(const UniValue& params, bool fHelp)
     uint256 blockHash;
     if (params.size() == 2)
     {
-        std::string strBlockHash = params[1].get_str();
-        blockHash = uint256S(strBlockHash);
+        uint256 hash = ParseHashV(params[1], "blockhash (optional parameter 2)");
```
kaykurokawa: This needs to set blockHash, otherwise its just setting a variable that never gets used

src/rpc/claimtrie.cpp
kaykurokawa: It's not wrong, but also not quite right.
ParseHashV is for hex strings representing total 256 bits for txid's.
ClaimId's are 160 bit so is represented by 40 hex chars. 
I would prefer you create a function similar to ParseHashV in rpc/claimtrie.cpp that is specifically for ClaimId's, maybe called ParseClaimId()

src/test/rpc_tests.cpp
```diff
@@ -332,4 +332,17 @@ BOOST_AUTO_TEST_CASE(rpc_convert_values_generatetoaddress)
     BOOST_CHECK_EQUAL(result[2].get_int(), 9);
 }
 
+BOOST_AUTO_TEST_CASE(rpc_claimtrie_validation) {
+    // std::runtime_error: parameter 2 must be hexadecimal string (not 'not_hex')
+    BOOST_CHECK_THROW(CallRPC("getnameproof test not_hex"), runtime_error);
+    // std::runtime_error: Block not found
+    BOOST_CHECK_THROW(CallRPC("getnameproof test a3b807410df0b60fcb9736768df5823938b2f838694939ba45f3c0a1bff150ed"), runtime_error);
+
```
kaykurokawa: Nice tests!
can we add a test here where we give getnameproof a valid block hash and it succeeds? 
You should be able to use the "generate" RPC call here to generate a block that you can use the blockhash for.

src/rpc/claimtrie.cpp
winder: Would it be acceptable to use `ParseHexV` instead of creating a new method? It does return a vector so there would be an unnecessary allocation, but it does the right thing without writing new code.

src/test/rpc_tests.cpp
winder: Great, I was wondering about that! I'll give that a try.

src/rpc/claimtrie.cpp
kaykurokawa: No, I think it is not acceptable. If you enter in the wrong amount of characters, it should be able to catch that.

If you don't want to write it though, I can do it. Or maybe @tzarebczan should give you more tip for the work. Let me know.

winder: Ok, that makes sense. It looks like `ParseHashV` doesn't do length validation either, perhaps it should be updated as well to check that the length is 32?
```
uint256 ParseHashV(const UniValue& v, string strName)
{
    string strHex;
    if (v.isStr())
        strHex = v.get_str();
    if (!IsHex(strHex)) // Note: IsHex("") is false
        throw JSONRPCError(RPC_INVALID_PARAMETER, strName+" must be hexadecimal string (not '"+strHex+"')");
    uint256 result;
    result.SetHex(strHex);
    return result;
}
```

kaykurokawa: On the most recent Bitcoin Core code base has updated the ParseHashV function to check this: https://github.com/bitcoin/bitcoin/blob/master/src/rpc/server.cpp#L118

So no you do not need to update ParseHashV since we plan to merge with the most recent Bitcoin Core code base soon.

lbrynaut:
The style needs to be addressed (clang-format).

winder:
I wasn’t sure what to do about the failing builds. There error messages are
sort of hidden inside the reproducible build script.

kaykurokawa:
I will rebase/squash/fix the travis errors and merge. Thanks for the contribution!

kaykurokawa:
rebased/formatted/squashed and merged

Issue #119, opened 2018-04-16
Claim name returned is strange


kaykurokawa:
Via Lex: 

working on unit tests for the claim name script and using real live blockchain data and am confused about `lbrycrd` output:

        ```"asm": "OP_CLAIM_NAME 1937006947 080110011a7808011230080410011a084d616361726f6e6922002a003214416c6c207269676874732072657365727665642e38004a0052005a001a42080110011a30add80aaf02559ba09853636a0658c42b727cb5bb4ba8acedb4b7fe656065a47a31878dbf9912135ddb9e13806cc1479d220a696d6167652f6a7065672a5c080110031a404180cc0fa4d3839ee29cca866baed25fafb43fca1eb3b608ee889d351d3573d042c7b83e2e643db0d8e062a04e6e9ae6b90540a2f95fe28638d0f18af4361a1c2214f73de93f4299fb32c32f949e02198a8e91101abd OP_2DROP OP_DROP OP_DUP OP_HASH160 be16e4b0f9bd8f6d47d02b3a887049c36d3b84cb OP_EQUALVERIFY OP_CHECKSIG",
        "hex": "b504636174734cdc080110011a7808011230080410011a084d616361726f6e6922002a003214416c6c207269676874732072657365727665642e38004a0052005a001a42080110011a30add80aaf02559ba09853636a0658c42b727cb5bb4ba8acedb4b7fe656065a47a31878dbf9912135ddb9e13806cc1479d220a696d6167652f6a7065672a5c080110031a404180cc0fa4d3839ee29cca866baed25fafb43fca1eb3b608ee889d351d3573d042c7b83e2e643db0d8e062a04e6e9ae6b90540a2f95fe28638d0f18af4361a1c2214f73de93f4299fb32c32f949e02198a8e91101abd6d7576a914be16e4b0f9bd8f6d47d02b3a887049c36d3b84cb88ac",```

looking at the hex: `b5` is correct for `OP_CLAIM_NAME`, then it's `04` which is correct for length of upcoming string, consuming the next 4 bytes (8 in hex) gives us `cats`:

```>>> '63617473'.decode('hex')
'cats'```

so... wtf is `1937006947` as returned by lbrycrd as the claim name?
it's not hex and it's not ascii



### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
[...]

PR #172, opened 2018-07-13
Claim name returned is strange

mirgee:
Fixing issue #119 .

In bitcoin, small stack pushes are treated as integers when converting scripts to asm. For our use case, it is more sensible to print all pushes as hex strings.

Using tx hex string from the issue thread, tested with 

```./lbrycrd-cli decodescript b504636174734cdc080110011a7808011230080410011a084d616361726f6e6922002a003214416c6c207269676874732072657365727665642e38004a0052005a001a42080110011a30add80aaf02559ba09853636a0658c42b727cb5bb4ba8acedb4b7fe656065a47a31878dbf9912135ddb9e13806cc1479d220a696d6167652f6a7065672a5c080110031a404180cc0fa4d3839ee29cca866baed25fafb43fca1eb3b608ee889d351d3573d042c7b83e2e643db0d8e062a04e6e9ae6b90540a2f95fe28638d0f18af4361a1c2214f73de93f4299fb32c32f949e02198a8e91101abd6d7576a914be16e4b0f9bd8f6d47d02b3a887049c36d3b84cb88ac```

which gives

```
{
  "asm": "OP_CLAIM_NAME 63617473 080110011a780801123...",
   ...
}
```

and this is the expected behavior, since `63617473` is hex for `cats`.

I will follow with further testing or unit tests if necessary.

### DIFF
--- a/src/core_write.cpp
+++ b/src/core_write.cpp
@@ -87,27 +87,23 @@ string ScriptToAsmStr(const CScript& script, const bool fAttemptSighashDecode)
             return str;
         }
         if (0 <= opcode && opcode <= OP_PUSHDATA4) {
-            if (vch.size() <= static_cast<vector<unsigned char>::size_type>(4)) {
-                str += strprintf("%d", CScriptNum(vch, false).getint());
-            } else {
-                // the IsUnspendable check makes sure not to try to decode OP_RETURN data that may match the format of a signature
-                if (fAttemptSighashDecode && !script.IsUnspendable()) {
-                    string strSigHashDecode;
-                    // goal: only attempt to decode a defined sighash type from data that looks like a signature within a scriptSig.
-                    // this won't decode correctly formatted public keys in Pubkey or Multisig scripts due to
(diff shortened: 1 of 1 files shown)


BrannonKing:
Are you able to add a unit test for this scenario?

lbrynaut:
@BrannonKing Has it been determined that this "fix" is needed?

kaykurokawa:
I was thinking that perhaps this might be too broad of a fix, and it would be better to detect if a script contains a claim OP code and handle it differently, rather than applying this fix to all scripts. Some application might depend on this particular behavior for decoderawtransaction and we might end up breaking it.. although I don't know of any application that does.

lbrynaut:
@kaykurokawa Sounds reasonable.  I think we should revisit this after upstream changes because 1) It's not urgent, and 2) Arguably it's not broken (although perhaps not easily understood).

kaykurokawa:
Closing this. Just noti

PR #175, opened 2018-07-19
Separate disk related functions in CClaimTrieDb

kaykurokawa:
This is a rebased and squashed version of @bvbfan 's CClaimTrieDb commits which separates disk related function from the claimtrie into its own class (from https://github.com/lbryio/lbrycrd/pull/160)

I removed here commits which renamed CClaimTrieCache and nCurrentHeight, as well as the new management of CClaimTrieNode, they will be put into seperate PR's. 

This PR is not downgrade compatible (once you upgrade to this commit, you will not be able to downgrade to previous versions without wiping your claim database) and should be accompanied by a version bump.

### DIFF
--- a/src/claimtrie.cpp
+++ b/src/claimtrie.cpp
@@ -2,9 +2,13 @@
 #include "coins.h"
 #include "hash.h"
 
-#include <boost/scoped_ptr.hpp>
-#include <iostream>
 #include <algorithm>
+#include <iostream>
+
+#define HASH_BLOCK 'h'
+#define CURRENT_HEIGHT 't'
+
+const uint256 one = uint256S("0000000000000000000000000000000000000000000000000000000000000001");
 
 std::vector<unsigned char> heightToVch(int n)
 {
@@ -21,7 +25,7 @@ std::vector<unsigned char> heightToVch(int n)
     return vchHeight;
 }
 
-uint256 getValueHash(COutPoint outPoint, int nHeightOfLastTakeover)
+uint256 getValueHash(const COutPoint& outPoint, int nHeightOfLastTakeover)
 {
     CHash256 txHasher;
     txHasher.Write(outPoint.hash.begin(), outPoint.hash.size());
@@ -32,7 +36,7 @@ uint256 getValueHash(COutPoint outPoint, int nHeightOfLastTakeover)
     std::stringstream ss;
     ss << outPoint.n;
     std::string snOut = ss.str();
-    nOutHasher.Write((unsigned char*) snOut.data(), snOut.size());
+    nOutHasher.Write((unsigned char*)snOut.data(), snOut.size());
     std::vector<unsigned char> vchnOutHash(nOutHasher.OUTPUT_SIZE);
     nOutHasher.Finalize(&(vchnOutHash[0]));
 
@@ -53,7 +57,7 @@ uint256 getValueHash(COutPoint outPoint, int nHeightOfLastTakeover)
     return valueHash;
 }
 
-bool CClaimTrieNode::insertClaim(CClaimValue claim)
+bool CClaimTrieNode::insertClaim(const CClaimValue& claim)
 {
     LogPrintf("%s: Inserting %s:%d (amount: %d)  into the claim trie\n", __func__, claim.outPoint.hash.ToString(), claim.outPoint.n, claim.nAmount);
     claims.push_back(claim);
@@ -65,24 +69,18 @@ bool CClaimTrieNode::removeClaim(const COutPoint& outPoint, CClaimValue& claim)
     LogPrintf("%s: Removing txid: %s, nOut: %d from the claim trie\n", __func__, outPoint.hash.ToString(), outPoint.n);
 
     std::vector<CClaimValue>::iterator itClaims;
-    for (itClaims = claims.begin(); itClaims != claims.end(); ++itClaims)
-    {
-        if (itClaims->outPoint == outPoint)
-        {
+    for (itClaims = claims.begin(); itClaims != claims.end(); ++itClaims) {
+        if (itClaims->outPoint == outPoint) {
             std::swap(claim, *itClaims);
             break;
         }
     }
-    if (itClaims != claims.end())
-    {
+    if (itClaims != claims.end()) {
         claims.erase(itClaims);
-    }
-    else
-    {
+    } else {
         LogPrintf("CClaimTrieNode::%s() : asked to remove a claim that doesn't exist\n", __func__);
         LogPrintf("CClaimTrieNode::%s() : claims that do exist:\n", __func__);
-        for (unsigned int i = 0; i < claims.size(); i++)
-        {
+        for (unsigned int i = 0; i < claims.size(); i++) {
             LogPrintf("\ttxhash: %s, nOut: %d:\n", claims[i].outPoint.hash.ToString(), claims[i].outPoint.n);
         }
         return false;
@@ -92,12 +90,9 @@ bool CClaimTrieNode::removeClaim(const COutPoint& outPoint, CClaimValue& claim)
 
 bool CClaimTrieNode::getBestClaim(CClaimValue& claim) const
 {
-    if (claims.empty())
-    {
+    if (claims.empty()) {
         return false;
-    }
-    else
-    {
+    } else {
         claim = claims.front();
         return true;
     }
@@ -105,10 +100,8 @@ bool CClaimTrieNode::getBestClaim(CClaimValue& claim) const
 
 bool CClaimTrieNode::haveClaim(const COutPoint& outPoint) const
 {
-    for (std::vector<CClaimValue>::const_iterator itclaim = claims.begin(); itclaim != claims.end(); ++itclaim)
-    {
-        if (itclaim->outPoint == outPoint)
-            return true;
+    for (std::vector<CClaimValue>::const_iterator itclaim = claims.begin(); itclaim != claims.end(); ++itclaim) {
+        if (itclaim->outPoint == outPoint) return true;
     }
     return false;
 }
@@ -117,17 +110,13 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
 {
     std::vector<CClaimValue>::iterator itclaim;
 
-    for (itclaim = claims.begin(); itclaim != claims.end(); ++itclaim)
-    {
+    for (itclaim = claims.begin(); itclaim != claims.end(); ++itclaim) {
         itclaim->nEffectiveAmount = itclaim->nAmount;
     }
 
-    for (supportMapEntryType::iterator itsupport = supports.begin(); itsupport != supports.end(); ++itsupport)
-    {
-        for (itclaim = claims.begin(); itclaim != claims.end(); ++itclaim)
-        {
-            if (itsupport->supportedClaimId == itclaim->claimId)
-            {
+    for (supportMapEntryType::iterator itsupport = supports.begin(); itsupport != supports.end(); ++itsupport) {
+        for (itclaim = claims.begin(); itclaim != claims.end(); ++itclaim) {
+            if (itsupport->supportedClaimId == itclaim->claimId) {
                 itclaim->nEffectiveAmount += itsupport->nAmount;
                 break;
             }
@@ -135,6 +124,20 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
     }
 
     std::make_heap(claims.begin(), claims.end());
+}
+
+bool CClaimTrieNode::empty() const
+{
+    return children.empty() && claims.empty();
+}
+
+CClaimTrie::CClaimTrie(bool fMemory, bool fWipe, int nProportionalDelayFactor)
+    : nCurrentHeight(0), nExpirationTime(Params().GetConsensus().nOriginalClaimExpirationTime), nProportionalDelayFactor(nProportionalDelayFactor), db(fMemory, fWipe), root(one)
+{
+}
+
+CClaimTrie::~CClaimTrie()
+{
 }
 
 uint256 CClaimTrie::getMerkleHash()
@@ -147,72 +150,24 @@ bool CClaimTrie::empty() const
     return root.empty();
 }
 
-template<typename K> bool CClaimTrie::keyTypeEmpty(char keyType, K& dummy) const
-{
-    boost::scoped_ptr<CDBIterator> pcursor(const_cast<CDBWrapper*>(&db)->NewIterator());
-    pcursor->SeekToFirst();
-    
-    while (pcursor->Valid())
-    {
-        std::pair<char, K> key;
-        if (pcursor->GetKey(key))
-        {
-            if (key.first == keyType)
-            {
-                return false;
-            }
-        }
-        else
-        {
-            break;
-        }
-        pcursor->Next();
-    }
-    return true;
-}
-
 bool CClaimTrie::queueEmpty() const
 {
-    for (claimQueueType::const_iterator itRow = dirtyQueueRows.begin(); itRow != dirtyQueueRows.end(); ++itRow)
-    {
-        if (!itRow->second.empty())
-            return false;
-    }
-    int dummy;
-    return keyTypeEmpty(CLAIM_QUEUE_ROW, dummy);
+    return db.keyTypeEmpty<int, claimQueueRowType>();
 }
 
 bool CClaimTrie::expirationQueueEmpty() const
 {
-    for (expirationQueueType::const_iterator itRow = dirtyExpirationQueueRows.begin(); itRow != dirtyExpirationQueueRows.end(); ++itRow)
-    {
-        if (!itRow->second.empty())
-            return false;
-    }
-    int dummy;
-    return keyTypeEmpty(EXP_QUEUE_ROW, dummy);
+    return db.keyTypeEmpty<int, expirationQueueRowType>();
 }
 
 bool CClaimTrie::supportEmpty() const
 {
-    for (supportMapType::const_iterator itNode = dirtySupportNodes.begin(); itNode != dirtySupportNodes.end(); ++itNode)
-    {
-        if (!itNode->second.empty())
-            return false;
-    }
-    std::string dummy;
-    return keyTypeEmpty(SUPPORT, dummy);
+    return db.keyTypeEmpty<std::string, supportMapEntryType>();
 }
 
 bool CClaimTrie::supportQueueEmpty() const
 {
-    for (supportQueueType::const_iterator itRow = dirtySupportQueueRows.begin(); itRow != dirtySupportQueueRows.end(); ++itRow)
-    {
-        if (!itRow->second.empty())
-            return false;
-    }
-    int dummy;
-    return keyTypeEmpty(SUPPORT_QUEUE_ROW, dummy);
+    return db.keyTypeEmpty<int, supportQueueRowType>();
 }
 
 void CClaimTrie::setExpirationTime(int t)
@@ -228,37 +183,25 @@ void CClaimTrie::clear()
 
 void CClaimTrie::clear(CClaimTrieNode* current)
 {
-    for (nodeMapType::const_iterator itchildren = current->children.begin(); itchildren != current->children.end(); ++itchildren)
-    {
+    for (nodeMapType::const_iterator itchildren = current->children.begin(); itchildren != current->children.end(); ++itchildren) {
         clear(itchildren->second);
         delete itchildren->second;
     }
+    current->children.clear();
 }
 
 bool CClaimTrie::haveClaim(const std::string& name, const COutPoint& outPoint) const
 {
-    const CClaimTrieNode* current = &root;
-    for (std::string::const_iterator itname = name.begin(); itname != name.end(); ++itname)
-    {
-        nodeMapType::const_iterator itchildren = current->children.find(*itname);
-        if (itchildren == current->children.end())
-            return false;
-        current = itchildren->second;
-    }
-    return current->haveClaim(outPoint);
+    const CClaimTrieNode* current = getNodeForName(name);
+    return current && current->haveClaim(outPoint);
 }
 
 bool CClaimTrie::haveSupport(const std::string& name, const COutPoint& outPoint) const
 {
     supportMapEntryType node;
-    if (!getSupportNode(name, node))
-    {
-        return false;
-    }
-    for (supportMapEntryType::const_iterator itnode = node.begin(); itnode != node.end(); ++itnode)
-    {
-        if (itnode->outPoint == outPoint)
-            return true;
+    if (!db.getQueueRow(name, node)) return false;
+    for (supportMapEntryType::const_iterator itnode = node.begin(); itnode != node.end(); ++itnode) {
+        if (itnode->outPoint == outPoint) return true;
     }
     return false;
 }
@@ -266,32 +209,20 @@ bool CClaimTrie::haveSupport(const std::string& name, const COutPoint& outPoint)
 bool CClaimTrie::haveClaimInQueue(const std::string& name, const COutPoint& outPoint, int& nValidAtHeight) const
 {
     queueNameRowType nameRow;
-    if (!getQueueNameRow(name, nameRow))
-    {
-        return false;
-    }
+    if (!db.getQueueRow(name, nameRow)) return false;
     queueNameRowType::const_iterator itNameRow;
-    for (itNameRow = nameRow.begin(); itNameRow != nameRow.end(); ++itNameRow)
-    {
-        if (itNameRow->outPoint == outPoint)
-        {
+    for (itNameRow = nameRow.begin(); itNameRow != nameRow.end(); ++itNameRow) {
+        if (itNameRow->outPoint == outPoint) {
             nValidAtHeight = itNameRow->nHeight;
             break;
         }
     }
-    if (itNameRow == nameRow.end())
-    {
-        return false;
-    }
+    if (itNameRow == nameRow.end()) return false;
     claimQueueRowType row;
-    if (getQueueRow(nValidAtHeight, row))
-    {
-        for (claimQueueRowType::const_iterator itRow = row.begin(); itRow != row.end(); ++itRow)
-        {
-            if (itRow->first == name && itRow->second.outPoint == outPoint)
-            {
-                if (itRow->second.nValidAtHeight != nValidAtHeight)
-                {
+    if (db.getQueueRow(nValidAtHeight, row)) {
+        for (claimQueueRowType::const_iterator itRow = row.begin(); itRow != row.end(); ++itRow) {
+            if (itRow->first == name && itRow->second.outPoint == outPoint) {
+                if (itRow->second.nValidAtHeight != nValidAtHeight) {
                     LogPrintf("%s: An inconsistency was found in the claim queue. Please report this to the developers:\nDifferent nValidAtHeight between named queue and height queue\n: name: %s, txid: %s, nOut: %d, nValidAtHeight in named queue: %d, nValidAtHeight in height queue: %d current height: %d\n", __func__, name, outPoint.hash.GetHex(), outPoint.n, nValidAtHeight, itRow->second.nValidAtHeight, nCurrentHeight);
                 }
                 return true;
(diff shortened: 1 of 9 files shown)


src/claimtriedb.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include <vector>
+#include <iterator>
+#include <boost/scoped_ptr.hpp>
+
+/// Concrete implementation of dirty queue
+template <typename K, typename V>
+struct CCMap : public CCBase
+{
+    CCMap() : data() {}
+    typedef std::vector<std::pair<K, V> > dataType;
```
BrannonKing: Why not an unordered_map? (or map if ordering matters). It appears from the method below that we commonly need to find by key.

bvbfan: unordered_map is a hash_map, so we should provide a hash algorithm for some classes, like base_blob derived.

src/claimtrie.cpp
```diff
@@ -2,10 +2,16 @@
...
 #include "hash.h"
 
-#include <boost/scoped_ptr.hpp>
 #include <iostream>
 #include <algorithm>
 
+#define HASH_BLOCK 'h'
+#define CURRENT_HEIGHT 't'
+
+#include "claimtriedb.cpp" /// avoid linker problems
```
lbrynaut: No, this is broken.

src/claimtrie.cpp
```diff
@@ -238,11 +198,9 @@ void CClaimTrie::clear(CClaimTrieNode* current)
 bool CClaimTrie::haveClaim(const std::string& name, const COutPoint& outPoint) const
 {
     const CClaimTrieNode* current = &root;
-    for (std::string::const_iterator itname = name.begin(); itname != name.end(); ++itname)
-    {
+    for (std::string::const_iterator itname = name.begin(); itname != name.end(); ++itname) {
```
lbrynaut: If this style change (and others like it) isn't pulled in from clang-format, remove it.

src/claimtrie.cpp
```diff
@@ -238,11 +198,9 @@ void CClaimTrie::clear(CClaimTrieNode* current)
...
         nodeMapType::const_iterator itchildren = current->children.find(*itname);
-        if (itchildren == current->children.end())
-            return false;
+        if (itchildren == current->children.end()) return false;
```
lbrynaut: If this style change (and others like it) isn't pulled in from clang-format, remove it. (Please apply to other cases as well -- alternatively, just run clang-format and rebase and comment that that's why it's here.)

src/claimtrie.cpp
```diff
@@ -441,22 +368,22 @@ const CClaimTrieNode* CClaimTrie::getNodeForName(const std::string& name) const
 bool CClaimTrie::getInfoForName(const std::string& name, CClaimValue& claim) const
 {
     const CClaimTrieNode* current = getNodeForName(name);
-    if (current)
-    {
+    if (current) {
         return current->getBestClaim(claim);
+    } else {
```
lbrynaut: Adding this else isn't needed.

src/claimtrie.cpp
```diff
@@ -441,22 +368,22 @@ const CClaimTrieNode* CClaimTrie::getNodeForName(const std::string& name) const
...
 
 bool CClaimTrie::getLastTakeoverForName(const std::string& name, int& lastTakeoverHeight) const
 {
     const CClaimTrieNode* current = getNodeForName(name);
-    if (current && !current->claims.empty())
-    {
+    if (current && !current->claims.empty()) {
         lastTakeoverHeight = current->nHeightOfLastTakeover;
         return true;
+    } else {
```
lbrynaut: Again, unless it's clang-format specified, don't add lines that aren't needed.

src/claimtrie.cpp
```diff
@@ -465,56 +392,41 @@ claimsForNameType CClaimTrie::getClaimsForName(const std::string& name) const
     std::vector<CSupportValue> supports;
     int nLastTakeoverHeight = 0;
     const CClaimTrieNode* current = getNodeForName(name);
-    if (current)
-    {
-        if (!current->claims.empty())
-        {
+    if (current) {
+        if (!current->claims.empty()) {
```
lbrynaut: Suggested style: if (current && !current->claims.empty())

src/claimtrie.cpp
```diff
@@ -1634,156 +1096,107 @@ bool CClaimTrieCache::addClaimToQueues(const std::string& name, CClaimValue& cla
...
-            base->removeFromClaimIndex(claim);
-            return true;
-        }
     }
     LogPrintf("%s: An inconsistency was found in the claim queue. Please report this to the developers:\nFound in named queue but not in height queue: name: %s, txid: %s, nOut: %d, nValidAtHeight: %d, current height: %d\n", __func__, name, outPoint.hash.GetHex(), outPoint.n, itQueueName->nHeight, nCurrentHeight);
     return false;
 }
 
-bool CClaimTrieCache::undoAddClaim(const std::string& name, const COutPoint& outPoint, int nHeight) const
+bool CClaimTrieCache::undoAddClaim(const std::string& name, const COutPoint& outPoint, int nHeight)
```
lbrynaut: Why are we removing const from all of the methods?  Are we also removing mutable members?

src/claimtrie.cpp
```diff
@@ -963,164 +625,41 @@ bool CClaimTrie::updateTakeoverHeight(const std::string& name, int nTakeoverHeig
...
+    db.Write(HASH_BLOCK, hashBlock);
+    db.Write(CURRENT_HEIGHT, nCurrentHeight);
+    return db.Sync();
 }
 
 bool CClaimTrie::InsertFromDisk(const std::string& name, CClaimTrieNode* node)
 {
-    if (name.size() == 0)
-    {
+    if (name.size() == 0) {
```
lbrynaut: name.empty() ?

src/claimtrie.cpp
```diff
@@ -963,164 +625,41 @@ bool CClaimTrie::updateTakeoverHeight(const std::string& name, int nTakeoverHeig
...
-        {
-            batch.Write(std::make_pair(SUPPORT_QUEUE_NAME_ROW, itQueue->first), itQueue->second);
+        CClaimTrieNode *pNode = itcache->second;
+        uint32_t num_claims = pNode ? pNode->claims.size() : 0;
+        LogPrintf("%s: Writing %s to disk with %d claims\n", __func__, itcache->first, num_claims);
+        if (pNode) {
+            db.updateQueueRow(itcache->first, *pNode);
+        } else {
+            CClaimTrieNode emptyNode;
+            db.updateQueueRow(itcache->first, emptyNode);
```
lbrynaut: Alternative style:
```
CClaimTrieNode emptyNode;
db.updateQueueRow(itcache->first, pNode ? *pNode : emptyNode);
```

I'll stop commenting on style, perhaps it's pointless, or maybe clang-format will help (although not on things like this, which are adding unnecessary lines).

src/claimtrie.h
```diff
@@ -465,145 +468,134 @@ class CClaimTrieCache
...
-    mutable queueNameType claimQueueNameCache;
-    mutable expirationQueueType expirationQueueCache;
-    mutable supportMapType supportCache;
-    mutable supportQueueType supportQueueCache;
-    mutable queueNameType supportQueueNameCache;
-    mutable expirationQueueType supportExpirationQueueCache;
-    mutable std::set<std::string> namesToCheckForTakeover;
-    mutable std::map<std::string, int> cacheTakeoverHeights; 
-    mutable int nCurrentHeight; // Height of the block that is being worked on, which is
-                                // one greater than the height of the chain's tip
```
lbrynaut: This answers my previous question about removing mutable members.  Disregard.

src/claimtriedb.cpp
lbrynaut: I don't think the queues get to be that large, but perhaps someday it's worth measuring the performance.  As it is, you gain cache effects due to locality (I think this wins), but with a well defined hash and unordered_map you gain constant time retrieval.  This is not urgent for this PR and it's fine as-is.  I'd also advise (style) to just use std::find.std::find_if with a comparator if/when it's supported by the compiler (certainly after upstream merges).

src/claimtriedb.h
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * additional macros to complete hash algorithm
+ */
+#define  len(s)          (sizeof(s)/sizeof(*s)-1)
+#define   H1(s, i, x)    (x*65599u+(uint8_t)s[(i)<len(s)?len(s)-1-(i):len(s)])
+#define   H4(s, i, x)  H1(s, i,  H1(s, i+ 1,  H1(s, i+  2,  H1(s, i+  3, x))))
+#define  H16(s, i, x)  H4(s, i,  H4(s, i+ 4,  H4(s, i+  8,  H4(s, i+ 12, x))))
+#define  H64(s, i, x) H16(s, i, H16(s, i+16, H16(s, i+ 32, H16(s, i+ 48, x))))
+#define H256(s, i, x) H64(s, i, H64(s, i+64, H64(s, i+128, H64(s, i+192, x))))
+
```
lbrynaut: Where did these definitions come from?  It should be documented because 1) Not all hashes are equally appropriate, and 2) I'd hope we can replace most/all of these with boost:: or std:: hash implementations.  I don't think supporting these in our code forever is ideal.

src/claimtriedb.h
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * explicit template specialization,
+ * it should be provided where it's need a type hash
+ */
+#define HASH_VALUE(T1, T2) \
+template<> inline          \
+size_t hashType<T1, T2>()  \
+{                          \
+    return HASH(#T1 #T2);  \
+}
```
lbrynaut: If we really want to do this, there are plenty of template specialization varieties that can be made that resort to std::hash which eliminates the need for many of the custom hash variants.

src/claimtriedb.cpp
```diff
@@ -0,0 +1,204 @@
...
+    for (std::map<size_t, CCBase*>::iterator itQueue = queues.begin(); itQueue != queues.end(); ++itQueue) {
+        itQueue->second->write(itQueue->first, this);
+    }
+}
+
+template <typename K, typename V>
+bool CClaimTrieDb::getQueueRow(const K &key, V &row) const
+{
+    const size_t hash = hashType<K, V>();
+    std::map<size_t, CCBase*>::const_iterator itQueue = queues.find(hash);
```
lbrynaut: As a compromise (and simplicity), perhaps it's worth considering 2 step hashes such that the above hash == some_map[K][V]?  I see why you need these now, but we're jumping through some hoops to get there, no?  What I proposed is probably less space efficient, but I want to be sure we're doing this in a sustainable way going forward.

src/test/claimtriedb_tests.cpp
```diff
@@ -0,0 +1,127 @@
...
+    BOOST_CHECK(intclaimQueueRowType                == 2291382435);
+    BOOST_CHECK(stringCClaimTrieNode                == 2689964764);
+    BOOST_CHECK(intsupportQueueRowType              == 2519094245);
+    BOOST_CHECK(stringqueueNameRowType              == 2566197209);
+    BOOST_CHECK(stringsupportMapEntryType           == 212036296);
+    BOOST_CHECK(intexpirationQueueRowType           == 2977646682);
+    BOOST_CHECK(uint160CClaimIndexElement           == 1885673905);
+    BOOST_CHECK(stringsupportQueueNameRowType       == 389061475);
+    BOOST_CHECK(intsupportExpirationQueueRowType    == 3108630299);
+}
```
lbrynaut: I really wish we could avoid this kind of test, but if we go this route, at least it helps locate regressions.

src/claimtrie.cpp
lbrynaut: Update: Hopefully this has already been clang-formatting because a large portion of it is just mechanical style changes, and it's distracting for such a large commit.  I realize that as we're requiring clang-format now this should be less an issue in the future.  If these were added manually, I suggest an immediate format/rebase.

src/claimtrie.cpp
bvbfan: Yes.

src/claimtriedb.h
bvbfan: Get it from here -> http://lolengine.net/blog/2011/12/20/cpp-constant-string-hash it *should* be unique.

src/claimtriedb.h
bvbfan: Yes, it's helper macro. There is no need to write class names by hand.

src/test/claimtriedb_tests.cpp
bvbfan: The test aims to *guarantee* that all distinct type has an unique id, when new type is added unit test should be updated as well. The algorithm should produce same results on all different compilers and all its version, which is *not* guaranteed nor by std::type_id nor boost::type_id.

bvbfan:
Is there some stopper to merge my changes on master? I plan to rebase my last changes for wed meet.

src/claimtrie.cpp
BrannonKing: I agree that this is a bad pattern. This should technically cause linker problems due to double definition.

src/claimtrie.cpp
BrannonKing: I like the change. Saying that a removal method is `const` is misleading. The only place I would expect to see the `mutable` keyword is in an immutable object for the caching of expensive-to-compute fields.

src/claimtrie.cpp
bvbfan: It looks like but is not. Template has a external linking and it has few *hacks* to prevent problems:
1. Write template in class body - this is ugly and is not a proper C++ fix
2. Write a 3th file only with template implementation then include only. This adds an inconsistency between .h and .cpp with added .impl
3. Mark normal functions as inline *do* the trick and it's the natural way to keep consistency of .h and .cpp files.

lbrynaut: Including a cpp file from another means incorrect/bad organization.  That is what I meant by broken.

src/claimtrie.cpp
lbrynaut: This is produced from clang-format?

lbrynaut:
> Is there some stopper to merge my changes on master? I plan to rebase my last changes for wed meet.

This is under review.  Feel free to rebase, but it won't be merged (i.e. don't merge it) until review is complete.

src/claimtriedb.h
lbrynaut: This link didn't work for me.  It's also not really documenting why we want to use these hash values in our code base indefinitely, as opposed to re-organizing or sticking with existing boost/std ones (or some composition of).

src/test/claimtriedb_tests.cpp
lbrynaut: This is not a strong argument for why we want/need to support our own hashing here. This seems like a case of NIH, which should be avoided (https://en.wikipedia.org/wiki/Not_invented_here#In_computing).  I don't have an understanding of why we *must* go this route without considering other options.  What should be an organizational PR (as the name states separating methods out), looks to be far larger, including a breaking database re-organization which is something else entirely.

@bvbfan If you've had discussions about this work that I was not a part of (surely you have), I'd like to chat with whoever had them with you :-)  @kaykurokawa ?  If it's just a matter of me not being caught-up, I'll adjust the review accordingly.

bvbfan:
There is no problem to be reviewed but let be conceptional rather than a silly comments. I mean when you know better way to do it, describe it how to do.

lbrynaut:
> There is no problem to be reviewed but let be conceptional rather than a silly comments. I mean when you know better way to do it, describe it how to do.

What silly comment(s)?  I'm mostly asking questions on or pointing out things I don't know, or don't look right to me.  As you said, if this has already been reviewed despite my concerns, it'll be merged anyway.

I'm only reviewing it because I care about this code and wonder about future maintenance.  This work was not assigned to me, so I don't know a better way to do all of it (and haven't thought about it as much as you and/or @kaykurokawa).  If you've exhausted all alternative implementations and arrived here, it'll be merged and this is all disregarded anyway.  Since I don't know, I've added my 2 cents.  I don't know if that provides any value though.  2 "lbrynaut tokens" worth of input may not be worth much :-)

bvbfan:
Only for clarification -> "Including a cpp file from another means incorrect/bad organization. That is what I meant by broken." is a silly comment. This not broke anything and it's not forbidden or anything else.

lbrynaut:
> Only for clarification -> "Including a cpp file from another means incorrect/bad organization. That is what I meant by broken." is a silly comment. This not broke anything and it's not forbidden or anything else.

Of course it's not forbidden, but that doesn't mean that it's common or required either.  In fact, it would be a first for this entire code base, which is why it stands out.

bvbfan:
There is not problem to be implemented in other way, but i'm against pattern to described to be bad by default. About me, only better pattern can make other to be marked as bad.

lbrynaut:
> There is not problem to be implemented in other way, but i'm against pattern to described to be bad by default. About me, only better pattern can make other to be marked as bad.

Ok.  I don't agree though.  Most times I've seen cpp files included from others was due to bad organization and/or laziness.  Similar to "goto" statements.  There's nothing wrong with them, and often you can achieve a lot using them.  But I think they are also bad organizationally.  What I mean by organization in this sense affects readability and comprehension.  There are many ways to implement things that hide other issues (such as linker problems, or basic error checking), but that doesn't mean they're all 'good' to use, even if allowed.  In most cases, a slight re-organization can improve readability/flow, which also helps comprehension, which leads to better maintenance, etc.

We may not agree on style, and I'm ok with that.  But since we're both working on this code and this is not a common use for the project, I hesitate to introduce it just because we can.  My opinion on introducing this is "it's broken", meaning it absolutely should not be introduced.  But like I said, my token input may be worthless ;-)

bvbfan:
As i say before this is not a *regular* cpp file, it's like a template implementation file, since i'm not prefer to separate cpp and impl files i've made a *trick* to combine them.

src/claimtriedb.h
BrannonKing: @bvbfan , can you explain why these hash methods weren't necessary before but are now?

bvbfan: Sure. Whole idea behind these hashes is to use as safe as possible db. Till now they a predefined chars 'n', 'r' etc. which was used in claimtrie.cpp now we separated it and calls can be type safe e.g. you can only declare type and don't care what is predefined char, hash or anything else. The first idea was to use boost::type_id but for sake of not generate same hashes in all its versions (e.g. in 1.59 and 1.64 are not same) i aim to provide a proven algorithm to generate same hashes in all compilers and OS'es. That's why unit test matters to prove any type that we need has a unique value.

bvbfan: Not only this, generic functions prevent code duplication and it does not need to update db implementation every time when new type is introduced. I've notice many code duplication in claimtrie i aim to remove it at all, this kind of implementation do it.

kaykurokawa:
I think the linker problem where we have to include "claimtriedb.cpp" in claimtrie.cpp is just an unfortunate problem when working with template classes in C++. In one of my past project we put implementations in the header file to get around this. I believe you can also use forward declaration (not sure if that is the right term , see last solution here: http://www.cs.technion.ac.il/users/yechiel/c++-faq/separate-template-class-defn-from-decl.html ) but that looks like the worst solution in this case due to the number of functions and possible types. 

So I think in this case, the solution chosen here seems fine given that the other two solutions are equally bad. 

The point regarding the new hashing features @lbrynaut brought up is good. Maybe it was a bad idea to change the original simple hashing function (where each type was assigned a unique letter) to this one , considering that we could have avoided a database migration and the code would have been simpler. This is not @bvbvfan 's fault , I said that it would be OK to do this but in hindsight this was not the best idea (I can take responsibility and fix this ).

bvbfan:
I can propose a fix that will not introduce migration nor cpp include. It can be done by template implicit instantiation of every function like this:
`template<> inline
char hashType<std::string, CClaimTrieNode>()
{
    return 'n';
}
template CClaimTrieDb::getQueueRow<std::string, CClaimTrieNode>(); // explicit instantiation
`
This can be done in claimtrie.db so it will not needed to be include anywhere, when we add new type new explicit instantiation should be provided. I can do that on master and make a new RR if it accept by all of us.

kaykurokawa:
I think the above new solution looks good to me

lbrynaut:
> So I think in this case, the solution chosen here seems fine given that the other two solutions are equally bad.

I'm aware of the issue, but if we really have to go this route, I think it should be made clear in a more standard way.  For example, under the include dir having an impl directory with the implementation defined in .ipp files makes it clear what's going on at least.  It's true it's more or less the same thing, but projects like boost chose to do it that way, and a lot of projects have followed with good reason (a de facto standard).

That said, I still don't like the idea of needing these methods in the first place and would prefer to see the solution that does not require the new types at all nor the DB migration.  @kaykurokawa The one way migration makes me nervous because usually migrations are done for underlying reasons other than a migration for the sake of it.  Second, if the migration simplified the code or allowed us to eliminate unnecessary code, I'd be all on board -- but instead this is the opposite, taking on responsibility for more untested code that adds complexity.  If you see the need to go this route, let me know what's on your mind (I still think I'm missing some background on this).  Otherwise, I want to avoid all unnecessary code to avoid the future maintenance costs.

lbrynaut:
> I can propose a fix that will not introduce migration nor cpp include. It can be done by template implicit instantiation of every function like this:

I'd like to review this solution as a PR.

bvbfan:
There is no migrations or cpp/ipp include, see better. The worst case to me is separate directory with ipp files in.

bvbfan:
I push it in this branch, now it's able for review.

lbrynaut:
@bvbfan I discussed this briefly with @kaykurokawa and here is where I'm at with it:

1) This is hard to review because of all of the unnecessary style changes (as mentioned earlier, if this comes from clang-format, save formatting for a separate commit/PR, otherwise, this would benefit from reverting those).
2) I think this an over-engineered solution for the issue at hand, and I'm of the opinion that this is the kind of project that requires minimalistic changes in preference to large sweeping ones to accomplish a particular goal.  I'm assuming that you and @kaykurokawa had agreed on some scope of this work, so while it may exactly address what he had in mind, it's not for me to decide.
3) I see no urgency in merging this for any reason (functionally, etc).  While a re-organization would be nice, I don't see this as positively better than existing.  There's some benefit to breaking up the code into separate files, and some benefit to templatizing some common functionality, but as-is, this PR doesn't strike me as code we need or want to merge right now.
4) Instead of going around @kaykurokawa and re-defining the task or asking you to start over and simplify the scope, I'm noting that right now I'm focused on the normalization fork and the upstream code changes, so I'm punting on this and will revisit later if @kaykurokawa requests it -- otherwise, it's in his hands at the moment.

If anyone else thinks this work is urgently needed and @kaykurokawa is unable to add the required input, let me know and I'll re-prioritize.

bvbfan:
For me, existing code base is redundant and it contains duplicate code fragments, RR-s, not only this, solve issues, later ones memory leaks. So after all it's not an over-engineering at all, furthermore it simplifies all logic.

bvbfan:
Rebase to master, @kaykurokawa if you have a comments or advice please share them otherwise it ready to go in.

kaykurokawa:
Hey, this seems to have some problem syncing, seems to be a new problem that was not present in previous iterations of this PR.

After CClaimTrie::WriteToDisk() is called to write buffers to disk .. 

this problem occurs:

2018-08-18 22:25:59 incrementBlock: nCurrentHeight (before increment): 33913
2018-08-18 22:25:59 ERROR: ConnectBlock() : the merkle root of the claim trie does not match (actual=0000000000000000000000000000000000000000000000000000000000000001 vs block=c4827743b4a73ca1b35275805f0564b1fe6d294ea316199173d9039f41fa3f40)
2018-08-18 22:25:59 Misbehaving: 103.100.44.62:9246 (0 -> 100) BAN THRESHOLD EXCEEDE

It seems like the merkle root of the claim trie gets calculated the same as an empty trie after CClaimTrie::WriteToDisk() is called.

bvbfan:
You tests with some mock data, unit tests pass when i test them, maybe we should update them.

bvbfan:
Can you provide steps to reproduce?

bvbfan:
So i found what you mean i think, so
call WriteToDisk()
then not call ReadFromDisk
lead to not finding claims
I'll investigate if mean that.

bvbfan:
I found the problem
`diff --git a/src/claimtrie.cpp b/src/claimtrie.cpp
index e6d7250f..305455c2 100644
--- a/src/claimtrie.cpp
+++ b/src/claimtrie.cpp
@@ -630,7 +630,8 @@ bool CClaimTrie::WriteToDisk()
uint32_t num_claims = pNode ? pNode->claims.size() : 0;
LogPrintf("%s: Writing %s to disk with %d claims\n", __func__, itcache->first, num_claims);
if (pNode) {
-            db.updateQueueRow(itcache->first, *pNode);
+            CClaimTrieNode copy(*pNode);
+            db.updateQueueRow(itcache->first, copy);
} else {
CClaimTrieNode emptyNode;
db.updateQueueRow(itcache->first, emptyNode);`
void CClaimTrieDb::updateQueueRow makes a swap on IN parameter (i don't want to change this behavior but it's not used on node) witch cause root to became empty, so with patch above issue is fixed and ready for land.

src/claimtriedb.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+/**
+ * function to provide unique char of 2 types
+ */
+template <class T1, class T2>
+unsigned char hashType();
+
+/// Concrete implementation of buffer
+template <typename K, typename V>
+struct CActualBuffer : public CAbstractBuffer
```
BrannonKing: What if this inherited from std::map instead? You wouldn't need the find and insert methods below. You wouldn't need the dataType stuff. It seems like that would be less code.

src/claimtriedb.cpp
```diff
@@ -0,0 +1,215 @@
...
+    if (itQueue == queues.end()) {
+        itQueue = queues.insert(itQueue, std::pair<unsigned char, CAbstractBuffer*>(hash, new CActualBuffer<K, V>));
+    }
+    CActualBuffer<K, V>* map = static_cast<CActualBuffer<K, V>*>(itQueue->second);
+    typename CActualBuffer<K, V>::dataType::iterator itData = map->find(key);
+    if (itData == map->end()) {
+        itData = map->push_back(std::make_pair(key, V()));
+    }
+    std::swap(itData->second, row);
+}
```
BrannonKing: If you were using a real map here, the above four lines would turn into a single `map->at(key) = row;`. That seems more straightforward to me. Are you intentionally trying to return the old row here? If so, it should use the method return. The way that's coded is going to lead to unintended consequences.

src/claimtrie.cpp
```diff
@@ -963,164 +624,41 @@ bool CClaimTrie::updateTakeoverHeight(const std::string& name, int nTakeoverHeig
...
-    {
+    for (std::string::const_iterator itname = name.begin(); itname + 1 != name.end(); ++itname) {
         nodeMapType::iterator itchild = current->children.find(*itname);
-        if (itchild == current->children.end())
-            return false;
+        if (itchild == current->children.end()) return false;
         current = itchild->second;
     }
-    current->children[name[name.size()-1]] = node;
+    current->children[name[name.size() - 1]] = new CClaimTrieNode(*node);
```
BrannonKing: Why do we need to clone this? That looks like a good way to make a memory leak.

src/claimtrie.cpp
```diff
@@ -1434,198 +935,158 @@ bool CClaimTrieCache::removeClaimFromTrie(const std::string& name, const COutPoi
...
-    else
-    {
+    } else {
         return addClaimToQueues(name, claim);
     }
 }
 
-bool CClaimTrieCache::addClaimToQueues(const std::string& name, CClaimValue& claim) const
+template <typename K, typename V>
+typename std::map<K, V>::iterator CClaimTrieCache::getQueueCacheRow(const K& key, std::map<K, V>& map, bool createIfNotExists)
```
BrannonKing: I think this method needs to return a read-only collection. Are people going to add to modify that data thinking that they're updating the actual queue? In looking at the code, it appears that we do return the real buffer, but that's only one of three cases. The other two: we read something off disk or we use that `createIfNotExists` flag.

src/claimtrie.cpp
```diff
@@ -1634,156 +1095,107 @@ bool CClaimTrieCache::addClaimToQueues(const std::string& name, CClaimValue& cla
...
-        for (itQueue = itQueueRow->second.begin(); itQueue != itQueueRow->second.end(); ++itQueue)
-        {
-            if (name == itQueue->first && itQueue->second.outPoint == outPoint)
-            {
-                break;
+        for (itQueue = itQueueRow->second.begin(); itQueue != itQueueRow->second.end(); ++itQueue) {
+            if (name == itQueue->first && itQueue->second.outPoint == outPoint) {
+                std::swap(claim, itQueue->second);
+                itQueueNameRow->second.erase(itQueueName);
+                itQueueRow->second.erase(itQueue);
```
BrannonKing: What does this line do if getQueueCacheRow actually called Read?

src/claimtrie.cpp
```diff
@@ -2527,79 +1756,107 @@ bool CClaimTrieCache::clear() const
...
+            return false;
+        }
+    }
+    for (std::map<std::string, int>::iterator itheight = cacheTakeoverHeights.begin(); itheight != cacheTakeoverHeights.end(); ++itheight) {
+        if (!base->updateTakeoverHeight(itheight->first, itheight->second)) {
+            LogPrintf("%s: Failed to update takeover height for:%s\n", __func__, itheight->first);
+            return false;
+        }
+    }
+    update(claimQueueCache);
```
BrannonKing: Any way we can do these updates in a `for` loop instead? We have all the buffers in a collection, no?

src/claimtriedb.cpp
bvbfan: Map is ordered i.e. it'll reorder every insert, list benefits speed over it O(log N) vs O(N)

src/claimtriedb.cpp
bvbfan: Yeah, but i don't think it's worst, no?

src/claimtrie.cpp
bvbfan: It should, i'll provide memleak free RR next to this.

src/claimtrie.cpp
bvbfan: It private one, so it safe to return iterator over second argument, which is map.

src/claimtrie.cpp
bvbfan: CClaimTrieCache::update has a for loop.

src/claimtriedb.cpp
BrannonKing: Your current use of a list is O(1) insertion time; it's the lookup time that suffers. Once we get the upstream c++11, we could use an unordered_list, which is definitely faster. However, the number of buffers we have is small; none of these solutions are going to be a performance problem. Hence, I prefer the "less code" option.

src/claimtriedb.cpp
BrannonKing: I do think it's pretty bad; it's what caused the problem that Kay found, right?

src/claimtrie.cpp
BrannonKing: CDBWrapper::Read does not automatically add to the cache as far as I can tell.

src/claimtriedb.cpp
bvbfan: I don't have good reason for swap, but follows current behavior.

src/claimtrie.cpp
bvbfan: getQueueCacheRow put it to cache, line:1079, this is safe, it ease from cache.

bvbfan:
Fix clang issue.

bvbfan:
Rebase to resolve conflicts.

BrannonKing:
@lbrynaut has requested that we hold on this until his upstream merge changes come in.

PR #177, opened 2018-07-23
Support for CLion Intellisense

BrannonKing:
(not to be used for a full compile of the project)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,56 @@
+cmake_minimum_required(VERSION 3.7)
+project(lbrycrd_clion) # not expecting a full compile -- just something that allows clion syntax checking
+
+set (CMAKE_CXX_STANDARD 98) # 03 is not supported by cmake, but this will soon be 11 (after upstream bitcoin merge)
+# I thought that setting the standard would disable the clang-tidy c++11 tips; nope.
+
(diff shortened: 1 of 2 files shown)


lbrynaut:
@BrannonKing I'm not sure why this is needed, but since it doesn't impact any existing code (and so long as it continues not to) and you're willing to maintain it, I don't see the harm.  Perhaps it makes more sense to be moved into contrib/devtools though?  If it must exist in the root directory,  could it be sym-linked in autogen.sh or something?

lbrynaut:
@BrannonKing Ok, makes sense.  Will it still be self-contained in a single file in the root, or would it need additional structure?

BrannonKing:
Some people do argue for a CMakeLists.txt file in each subfolder, which is a possible approach. For myself, though, I like to keep it all in one at the parent level; it's easier to navigate that way.

CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+cmake_minimum_required(VERSION 3.7)
+project(lbrycrd_clion) # not expecting a full compile -- just something that allows clion syntax checking
+
```
kaykurokawa: maybe make this more explicit, so people don't get confused what this file is for ?  
"Do not use for full compile. This is for CLion syntax checking only".

kaykurokawa:
a small nit, 
untested but looks good to me.

Issue #83, opened 2018-01-23
Lbrycrd claimtrie unit test improvements 


kaykurokawa:
Old tests for the claimtrie ( https://github.com/lbryio/lbrycrd/blob/master/src/test/claimtrie_tests.cpp ) should be merged with new tests which contains fixtures and functions to make the expansion of unit tests easier: https://github.com/lbryio/lbrycrd/blob/master/src/test/claimtriebranching_tests.cpp

The new tests could also be improved more, it is mainly testing for the best claim, and we should test other things in the claim trie such as the number of claims in the trie, check if the claimtrie root merkle hash is being updated, check that supports are present in the claim trie directly, etc..

BrannonKing:
done via #183

PR #181, opened 2018-07-26
Unify claimtrie tests, add some additional root hash checks

BrannonKing:
As part of the review, look for the comment "tx1MerkleHash doesn't match right here" and tell me if that's expected.


src/test/claimtriebranching_tests.cpp
```diff
@@ -457,7 +468,7 @@ BOOST_AUTO_TEST_CASE(claimtriebranching_spend_claim)
         check claim with more support wins
         check support delay 
 */
-BOOST_AUTO_TEST_CASE(claimtriebranching_support)
+BOOST_AUTO_TEST_CASE(support_test)
 {
```
lbrynaut: The previous test names were far more descriptive when tracking down logged errors.  It described module and the test name, not a name with "_test" appended (which is non-informative, given that all unit tests are tests).

lbrynaut:
If this isn't ready yet, consider adding WIP (for work in progress to the PR name).

lbrynaut:
Also, just a general question -- why are these being merged?  I liked having the older tests separate from the first, which can help find regressions (have actually found several while I was working on the normalization).

BrannonKing:
As for the "why" question, @kaykurokawa can say why he filed #83 . There is 200 lines less code than there was before the test merge, and the old tests are still covering everything that they did before. It's just not so obvious now that they're older.

BrannonKing:
I think I finally got the right format command figured out. I'll push up that commit to make sure it builds on Travis. I'm happy to squash this before we merge, but I think the different commits allow you to more easily see what was added to test hashes and tree counts.

lbrynaut:
> I think I finally got the right format command figured out. I'll push up that commit to make sure it builds on Travis. I'm happy to squash this before we merge, but I think the different commits allow you to more easily see what was added to test hashes and tree counts.

Everyone has different opinions here, but my take is that pull requests with many commits are just distracting since either the pull request addresses the issue or not.  Most times when you see a stack of commits, they are exactly things like "ran formatting tool", "updated test", "something else useless", etc.  In reality, reviewers look at the change-set as a whole because the PR is more or less all or nothing.  Cherry-picking individual commits from a single PR is very rare -- I'd dare say it never happens.  Instead you see even more garbage commits like "updated test", " fixed the issue added by test", "reverted last change", "really updated test".

Developers tend to overvalue the sequence of how code arrived at a particular place -- for example, a common argument I hear for not squashing is that so that a particular developer can remember the process of making the changes, or some other sentimental/historical reason.  The problem is that anyone can do that in a local tree, but it's meaningless (and distracting) to other developers, the ones reviewing and/or maintaining it -- who are not interested in that sequence, and it actually makes history tracking somewhat more difficult because instead of a concrete place where PR X was merged (which can be reviewed/edited/undone), you have to potentially suffer through reviewing 40+ commits to find the actual spot where something happened/broke/changed (including all the 10+ other commits in there where it was twiddled back and forth for whatever reason during the 'process').

This is a bit of rant, not particularly just about your comment @BrannonKing.  I find clean source histories very important, particularly when we are looking to be tracking an upstream.  It's the entire reason that git is more powerful than other scms (the ability to hack/clean the timeline). I'm just advocating for professional standards on the main public repo.  Of course, anyone can do whatever they'd like on their personal repos.

@kaykurokawa and I largely agreed that squashing the commits was the least noisy (last we talked about it, although it's been a while).  If you find that this is not the right thing to do in some cases, just make the case. But keep the above in mind.  There's nothing wrong with formatting PRs, but they would ideally just be formatting -- not formatting and overhaul subsystem X at the same time.

src/test/claimtriebranching_tests.cpp
```diff
@@ -1053,15 +1060,1976 @@ BOOST_AUTO_TEST_CASE(claimtriebranching_hardfork_disktest)
...
+    std::string sValue3("123testa");
+
+    uint256 hash0(uint256S("0000000000000000000000000000000000000000000000000000000000000001"));
+    BOOST_CHECK(pclaimTrie->getMerkleHash() == hash0);
+
+
+    CMutableTransaction tx1 = BuildTransaction(fixture.GetCoinbase());
+    tx1.vout[0].scriptPubKey = CScript() << OP_CLAIM_NAME
+                                         << std::vector<unsigned char>(sName1.begin(), sName1.end())
+                                         << std::vector<unsigned char>(sValue1.begin(), sValue2.end()) << OP_2DROP << OP_DROP << OP_TRUE;
```
BrannonKing: This mismatch on the end is what was causing the Travis-CI test failure. Strange thing: it doesn't fail with the GCC 7 builds, with or without optimizations. Only the GCC 4 builds fail.

lbrynaut: Interesting.  If you're able to track down why that is, it may save us some headaches in the future.  Once we are inline with upstream, gcc 5/6/7/8 becomes a viable option, perhaps even a minimum supported version.

lbrynaut: This looks like a bug? ```std::vector<unsigned char>(sValue1.begin(), sValue2.end())```

lbrynaut: Also, does the construction form make any difference?

Changing something like:
```
std::vector<unsigned char>(sName1.begin(), sName1.end())
```

to

```
std::vector<unsigned char>(sName1.data(), sName1.data() + sName1.size())
```

lbrynaut: On second read ... are you just saying that the mismatch is/what the actual problem and later compilers don't notice it?  We can play with the warning levels if so later.

lbrynaut: Ah right, it would require a temporary, (I hacked it since we're not using c++11).  The preferred syntax (c++11) would be:

```
{ sName1.data(), sName1.data() + sName1.size() }
```

The test without is something like:

```
std::vector<unsigned char> foo(sName1.data(), sName1.data() + sName1.size())

...
<< foo << other_stuff;
```

lbrynaut: BTW, if the test passes with that bug, there's another bug at the surface.

BrannonKing: I didn't notice the `data()` on the first parameter previously. This does compile and also does not throw an error in GCC7: `... std::vector<unsigned char>(sValue1.data(), sValue2.data() + sValue2.size()) ...`. It probably works because sName1 and sName2 are in order on the stack; you end up with sValue1 prepended to sValue2 with some junk in the middle, but we never verify the value.

BrannonKing:
I believe that I now understand my question where the claimtrie merkle root hash didn't match: the height (er, block number) where the transaction is activated (known as node->nHeightOfLastTakeover) is also included in the hash input. The upper one is activated at height 141; the lower at height 153.

PR #182, opened 2018-07-30
Fixed spacing so OS paths show up on different lines

ocornoc:


tzarebczan:
Good catch, thanks for fixing this! Another trick in markdown is to use 3 spaces at the end of a line - will cause it to line break.

PR #183, opened 2018-07-31
moved claimtrie_tests into claimtriebranching_tests

BrannonKing:
... to take advantage of the test fixture there. I also added other checks to the tests for the claimtrie root hash and some checks for the current claim count.


kaykurokawa:
Looks good to Merge.  

I think one thing to consider though regarding the tests that were brought in from claimtrie_tests that are similar to existing ones in claimtriebranching_tests: 

insert_update_claim_test duplicates some tests on claimtrie_update_test and claim_test 

claim_expiration_test duplicates some tests on claimtrie_expire_test 

supporting_claims_test , supporting_claims_test_2 duplicates each other and some tests on support_test

I think combining these tests together so that there are less redundant tests would be nice. Would be nice to see in a separate PR but I think its not a huge priority.

PR #185, opened 2018-08-05
Specify non standard tx errors

bvbfan:

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1034,8 +1034,12 @@ bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state, const C
 
     // Rather not work on nonstandard transactions (unless -testnet/-regtest)
     string reason;
-    if (fRequireStandard && !IsStandardTx(tx, reason))
-        return state.DoS(0, false, REJECT_NONSTANDARD, reason);
+    if (fRequireStandard && !IsStandardTx(tx, reason)) {
+        if (reason == "dust")
+            return state.DoS(0, false, REJECT_DUST, reason);
+        else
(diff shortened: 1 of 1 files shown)


bvbfan:
To pass clang-format-diff.py should make an unrelated comment change in src/policy/policy.h

BrannonKing:
Was there an existing issue reported for this?

lbrynaut:
Just noting here that this is rather invasive (affecting consensus code; i.e. anything in consensus directory), and potentially breaks forward/upstream compatibility in perpetuity.  It's also another good example of over-engineering where not necessary.  IMO, the correct solution is to do what upstream did -- remove the unused constant.  Since it's just a matter of error reporting, replacing the perfectly good reason string with codes in this PR is complete overkill.

I don't agree with the comment from @kaykurokawa from the issue (and I suspect he would reconsider if he knew what upstream did), but even if I did, this PR should be maybe 2 lines long.

bvbfan:
@lbrynaut , @kaykurokawa i can simplify patch as
```
if (reason == "dust") 
    return state.DoS(0, false, REJECT_DUST, reason);
```
Other remains unchanged.

kaykurokawa:
I agree that this is a bit too much, which will affect upstream compatibility without a good reason. 
The stuff here is based on: https://github.com/bitcoin/bips/blob/master/bip-0061.mediawiki

I think the suggestion to add below line to AcceptToMemoryPoolWorker is fine:

```
if (reason == "dust") 
    return state.DoS(0, false, REJECT_DUST, reason);
```

bvbfan:
implement as suggest.

kaykurokawa:
LGTM

Issue #131, opened 2018-04-30
Output blockchain APIs in a format that can be used for documentation


kauffj:
lbry.tech will be used to showcase the API signatures for the daemon, blockchain, and any other key components.

kauffj:
@lyoshenka did this get discussed on blockchain weekly?

BrannonKing:
Waiting on #44

BrannonKing:
Is this a duplicate of #143 ?

lyoshenka:
yes. ill close the other one.

BrannonKing:
Also, see https://github.com/lbryio/lbry.tech/issues/42
and https://github.com/lbryio/lbry/blob/master/scripts/generate_json_api.py

I'm starting on this now and planning to pursue the jrgen plan.

rynomad:
@BrannonKing re: jrgen... if you'd like, take a look at [this script](https://github.com/rynomad/lbry-js/blob/master/scripts/jrgen_crd.js) I wrote for lbry-js; it's pretty barebones but generates a jrgen compatible schema file dynamically from the existing lbrycrd 'help' command.

BrannonKing:
@rynomad , thanks for posting that script. Unfortunately, my Python script was generally completed before I got this message yesterday. Hence, I think I'll proceed with my script. 

I was unable to run your script. I'm not sure I followed your instructions correctly. `npm --version` reports 3.5.2. Do I need some other version? (There was no `setup` command on mine.)

Issue #164, opened 2018-06-25
An automated test for checking that a PR can sync


kaykurokawa:
As part of the automated travis build (or maybe it could be a separate thing from travis) , we should try to run and sync the built lbrycrd on testnet to see if it can a) sync from scratch and b) start up when its already synced and can succesfuly recieve a couple of blocks.

Not sure if this is possible just using travis (might take too long? use too much disk space?). We might have to set up some servers for this.

lbrynaut:
Yeah, I agree it would be nice, but I think it would also require doing externally.  As a first cut, we could probably just grab the binaries built from the travis build, run them for however long it's trying to do things, then e-mail the last couple hundred lines of the log to determine if things were normal up until shutdown.

I think travis has timeout limits that would prevent it from working there.

lbrynaut:
Nothing wrong with regtest testing also, but what @kaykurokawa is suggesting is specifically testnet and mainnet, to ensure nothing obvious is overlooked in a PR.  Usually that kind of testing is done manually.

kaykurokawa:
It wouldn't hurt to have some quick tests to verify regtest within the test. 
Also, might also be good to try testing sync'ing to mainnet as well.

PR #187, opened 2018-08-08
added scripts to generate parsable API docs

BrannonKing:
Issue: #131 . There are two scripts: 
1. jrgen, matching https://github.com/mzernetsch/jrgen
2. "v1" which matches https://github.com/lbryio/lbry/blob/master/scripts/generate_json_api.py


eukreign:
Seems like there is a mix of py2 and py3 exclusive syntax used (`print` keyword in `contrib/devtools/clang-format.py` which will fail in py3 as syntax error and `def get_obj_from_dirty_text(full_object: str):` which will fail as syntax error in py2).

Is this intentional? Should all code be only py2, py3 or should it work in both py2 and py3?

BrannonKing:
@eukreign , the clang-format scripts predate my time here. I just made it run as it didn't before. I don't know that anyone here wrote them, judging from the copyright messages in them. As far as I know, they have always been Py2 scripts. For this PR, the focus should be on the "docs" scripts.

eukreign:
Alright, it was in this PR so I thought it was related. Which files should I review as part of this PR?

lyoshenka:
@BrannonKing can this be merged? also as part of this, can you run the script (the one that matches the daemon script) and commit the output as well?

BrannonKing:
Open questions on this:
1. Should we check in the generated output as well? (Anyone making changes to the API would be responsible to rerun the script and check in an updated file.)
2. Should we make the reproducible build script run the document generation?
3. Should we make TravisCI run the document generation? And tag changes? Or have it check in the output?

eukreign:
Following from my answer on slack about how to unit test these scripts: to be able to use the generated output as the test itself, you have to commit and version the output deliberately (not via travis) so that when the docs do change it can be verified for correctness (did the docs change because the input source changed or did they change because the generator script changed?). Therefore, my advice would be:
1. Yes.
2. Don't know what this means, something lbrycrd specific I assume?
3. If you have extra time to dedicate to this task then definitely it would be beneficial to run the generate script and verify that the latest output matches the previously commited output and fail the build if it doesn't (asking commiter to run the doc build script locally, verify the changes make sense and then commit the new json doc file).

src/rpc/claimtrie.cpp
```diff
@@ -382,16 +382,16 @@ UniValue getclaimbyid(const UniValue& params, bool fHelp)
                         "  \"n\"                   (numeric) vout value\n"
                         "  \"amount\"              (numeric) txout value\n"
                         "  \"effective amount\"    (numeric) txout amount plus amount from all supports associated with the claim\n"
-                        "  \"height\"              (numeric) the height of the block in which this claim transaction is located\n"
```
lbrynaut: Just curious why you're re-ordering here.  Is this related to something about the doc generation?

BrannonKing: "valid at height" was between the supports array declaration and its definition. That was definitely wrong. I moved "height" as well just so that it matched the code filling it in.

lbrynaut: :+1: Thanks

lbrynaut:
Looks good to me.  Has this been run through clang-format?  It seems travis is only unhappy with style.

BrannonKing:
I have not ran the formatting. I'm having to undo 20 years of habit to remember to format before checking in. I still want to make that a report rather than a requirement. 

It sounds like we would additionally like a report from Travis (or the local build) that states that the generated API doc needs to be updated. This involves starting the server. Have we done that on Travis before? Do we run any integration tests on Travis?

lbrynaut:
> Do we run any integration tests on Travis?

No.  Possibly related to #164

PR #189, opened 2018-08-10
added blockhash input to claimtrie RPC methods plus tests

BrannonKing:
I'm marking this as WIP as I'm unsure on the approach and looking for feedback. You'll notice that `getclaimsforname` is done differently (more robustly) than the others. The others could be done like that one but not vice-versa, but I'm less sure on the consequences of that approach and the completeness of what's there.

### DIFF
--- a/src/claimtrie.cpp
+++ b/src/claimtrie.cpp
@@ -1229,11 +1229,8 @@ uint256 CClaimTrieCache::getMerkleHash() const
     }
     if (dirty())
     {
-        nodeCacheType::iterator cachedNode = cache.find("");
-        if (cachedNode != cache.end())
-            recursiveComputeMerkleHash(cachedNode->second, "");
-        else
-            recursiveComputeMerkleHash(&(base->root), "");
+        CClaimTrieNode* root = getRoot();
+        recursiveComputeMerkleHash(root, "");
     }
     hashMapType::iterator ithash = cacheHashes.find("");
     if (ithash != cacheHashes.end())
@@ -1285,11 +1282,8 @@ bool CClaimTrieCache::getOriginalInfoForName(const std::string& name, CClaimValu
 bool CClaimTrieCache::insertClaimIntoTrie(const std::string& name, CClaimValue claim, bool fCheckTakeover) const
 {
     assert(base);
-    CClaimTrieNode* currentNode = &(base->root);
+    CClaimTrieNode* currentNode = getRoot();
     nodeCacheType::iterator cachedNode;
-    cachedNode = cache.find("");
-    if (cachedNode != cache.end())
-        currentNode = cachedNode->second;
     for (std::string::const_iterator itCur = name.begin(); itCur != name.end(); ++itCur)
     {
         std::string sCurrentSubstring(name.begin(), itCur);
@@ -1374,11 +1368,8 @@ bool CClaimTrieCache::insertClaimIntoTrie(const std::string& name, CClaimValue c
 bool CClaimTrieCache::removeClaimFromTrie(const std::string& name, const COutPoint& outPoint, CClaimValue& claim, bool fCheckTakeover) const
 {
     assert(base);
-    CClaimTrieNode* currentNode = &(base->root);
+    CClaimTrieNode* currentNode = getRoot();
     nodeCacheType::iterator cachedNode;
-    cachedNode = cache.find("");
-    if (cachedNode != cache.end())
-        currentNode = cachedNode->second;
     assert(currentNode != NULL); // If there is no root in either the trie or the cache, how can there be any names to remove?
     for (std::string::const_iterator itCur = name.begin(); itCur != name.end(); ++itCur)
     {
@@ -1448,10 +1439,7 @@ bool CClaimTrieCache::removeClaimFromTrie(const std::string& name, const COutPoi
         if (fCheckTakeover)
             namesToCheckForTakeover.insert(name);
     }
-    CClaimTrieNode* rootNode = &(base->root);
-    cachedNode = cache.find("");
-    if (cachedNode != cache.end())
-        rootNode = cachedNode->second;
+    CClaimTrieNode* rootNode = getRoot();
     return recursivePruneName(rootNode, 0, name);
 }
 
@@ -1752,12 +1740,7 @@ bool CClaimTrieCache::reorderTrieNode(const std::string& name, bool fCheckTakeov
     cachedNode = cache.find(name);
     if (cachedNode == cache.end())
     {
-        CClaimTrieNode* currentNode;
-        cachedNode = cache.find("");
-        if(cachedNode == cache.end())
-            currentNode = &(base->root);
-        else
-            currentNode = cachedNode->second;
+        CClaimTrieNode* currentNode = getRoot();
         for (std::string::const_iterator itCur = name.begin(); itCur != name.end(); ++itCur)
         {
             std::string sCurrentSubstring(name.begin(), itCur);
@@ -2575,7 +2558,7 @@ void CClaimTrieCache::recursiveFlattenTrie(const std::string& name, const CClaim
 std::vector<namedNodeType> CClaimTrieCache::flattenTrie() const
 {
     std::vector<namedNodeType> nodes;
-    recursiveFlattenTrie("", &(base->root), nodes);
+    recursiveFlattenTrie("", getRoot(), nodes);
     return nodes;
 }
 
@@ -2652,7 +2635,7 @@ bool CClaimTrieCache::getInfoForName(const std::string& name, CClaimValue& claim
     if (dirty())
         getMerkleHash();
 
-    CClaimTrieNode* current = &(base->root);
+    CClaimTrieNode* current = getRoot();
     nodeCacheType::const_iterator cachedNode;
 
     for (std::string::const_iterator itName = name.begin(); current; ++itName) {
@@ -2678,7 +2661,7 @@ bool CClaimTrieCache::getProofForName(const std::string& name, CClaimTrieProof& 
         getMerkleHash();
 
     std::vector<CClaimTrieProofNode> nodes;
-    CClaimTrieNode* current = &(base->root);
+    CClaimTrieNode* current = getRoot();
     nodeCacheType::const_iterator cachedNode;
     bool fNameHasValue = false;
     COutPoint outPoint;
--- a/src/claimtrie.cpp
+++ b/src/claimtrie.cpp
@@ -527,37 +527,24 @@ claimsForNameType CClaimTrie::getClaimsForName(const std::string& name) const
 }
 
 //return effective amount from claim, retuns 0 if claim is not found
-CAmount CClaimTrie::getEffectiveAmountForClaim(const std::string& name, uint160 claimId) const
+CAmount CClaimTrie::getEffectiveAmountForClaim(const std::string& name, const uint160& claimId, std::vector<CSupportValue>* supports) const
 {
-    std::vector<CSupportValue> supports;
-    return getEffectiveAmountForClaimWithSupports(name, claimId, supports);
+    return getEffectiveAmountForClaim(getClaimsForName(name), claimId, supports);
 }
 
-//return effective amount from claim and the supports used as inputs, retuns 0 if claim is not found
-CAmount CClaimTrie::getEffectiveAmountForClaimWithSupports(const std::string& name, uint160 claimId,
-                                                           std::vector<CSupportValue>& supports) const
+CAmount CClaimTrie::getEffectiveAmountForClaim(const claimsForNameType& claims, const uint160& claimId, std::vector<CSupportValue>* supports) const
 {
-    claimsForNameType claims = getClaimsForName(name);
     CAmount effectiveAmount = 0;
-    bool claim_found = false;
-    for (std::vector<CClaimValue>::iterator it=claims.claims.begin(); it!=claims.claims.end(); ++it)
-    {
-        if (it->claimId == claimId && it->nValidAtHeight < nCurrentHeight)
-        {
+    for (std::vector<CClaimValue>::const_iterator it = claims.claims.begin(); it != claims.claims.end(); ++it) {
+        if (it->claimId == claimId && it->nValidAtHeight < nCurrentHeight) {
             effectiveAmount += it->nAmount;
-            claim_found = true;
+            for (std::vector<CSupportValue>::const_iterator it = claims.supports.begin(); it != claims.supports.end(); ++it) {
+                if (it->supportedClaimId == claimId && it->nValidAtHeight < nCurrentHeight) {
+                    effectiveAmount += it->nAmount;
+                    if (supports) supports->push_back(*it);
+                }
+            }
             break;
-        }
-    }
-    if (!claim_found)
-        return effectiveAmount;
-
-    for (std::vector<CSupportValue>::iterator it=claims.supports.begin(); it!=claims.supports.end(); ++it)
-    {
-        if (it->supportedClaimId == claimId && it->nValidAtHeight < nCurrentHeight)
-        {
-            effectiveAmount += it->nAmount;
-            supports.push_back(*it);
         }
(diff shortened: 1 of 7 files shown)


BrannonKing:
Apparently I need to add those minconf parameters to the client.cpp file for automatic CLI conversion to int.

src/claimtrie.cpp
```diff
@@ -2554,10 +2557,75 @@ uint256 CClaimTrieCache::getLeafHashForProof(const std::string& currentPosition,
...
+        cachedNode = cache.find(str);
+        if (cachedNode != cache.end())
+            recursiveFlatten(str, cachedNode->second, nodes);
+        else
+            recursiveFlatten(str, it->second, nodes);
+    }
+}
+
+
+std::vector<namedNodeType> CClaimTrieCache::flattenTrie() const
```
BrannonKing: This method existed for CClaimTrie but not for the "cache".

src/claimtrie.cpp
```diff
@@ -2554,10 +2557,75 @@ uint256 CClaimTrieCache::getLeafHashForProof(const std::string& currentPosition,
...
+        if (cachedNode != cache.end())
+            recursiveFlatten(str, cachedNode->second, nodes);
+        else
+            recursiveFlatten(str, it->second, nodes);
+    }
+
+    return nodes;
+}
+
+bool CClaimTrieCache::getInfoForName(const std::string& name, CClaimValue& toBeFilled) const
```
BrannonKing: this method also existed for CClaimTrie but not for the cache; it was mostly duplicated here.

src/claimtrie.cpp
```diff
@@ -2575,9 +2643,10 @@ CClaimTrieProof CClaimTrieCache::getProofForName(const std::string& name) const
         uint256 valueHash;
         if (fNodeHasValue)
         {
-            int nHeightOfLastTakeover;
-            assert(getLastTakeoverForName(currentPosition, nHeightOfLastTakeover));
-            valueHash = getValueHash(claim.outPoint, nHeightOfLastTakeover);
+            int nHeightOfLastTakeover2;
+            if (!getLastTakeoverForName(currentPosition, nHeightOfLastTakeover2))
+                return false;
```
BrannonKing: I was making this follow the pattern of `getInfoForName`.

src/rpc/claimtrie.cpp
```diff
@@ -46,55 +48,54 @@ UniValue getclaimsintrie(const UniValue& params, bool fHelp)
...
-    CCoinsViewCache view(pcoinsTip);
-    std::vector<namedNodeType> nodes = pclaimTrie->flattenTrie();
+    int maxHeight = chainActive.Height();
+    if (params.size() > 0)
+        maxHeight -= params[0].get_int() - 1;
 
+    CCoinsViewCache coins(pcoinsTip);
+    CClaimTrieCache claimTrie(pclaimTrie);
+    if (maxHeight < chainActive.Height())
+        if (!RollBackTo(coins, claimTrie, chainActive[maxHeight]))
```
BrannonKing: This method uses the approach of rolling back the caches to put us in a state where we were some number of blocks back. The caches are never flushed.

src/rpc/claimtrie.cpp
```diff
@@ -197,19 +209,32 @@ UniValue getvalueforname(const UniValue& params, bool fHelp)
...
+    if (!getValueForClaim(coins, claim.outPoint, sValue))
         return ret;
+
     ret.push_back(Pair("value", sValue));
     ret.push_back(Pair("claimId", claim.claimId.GetHex()));
     ret.push_back(Pair("txid", claim.outPoint.hash.GetHex()));
     ret.push_back(Pair("n", (int)claim.outPoint.n));
     ret.push_back(Pair("amount", claim.nAmount));
-    ret.push_back(Pair("effective amount", pclaimTrie->getEffectiveAmountForClaim(name, claim.claimId))); 
+    ret.push_back(Pair("effective amount", pclaimTrie->getEffectiveAmountForClaim(name, claim.claimId, maxHeight)));
```
BrannonKing: This particular line is not using data from the caches. It's hitting the live claimTrie itself. It appears that passing in the maxHeight would give the same result as rolling back the main trie.

src/rpc/claimtrie.cpp
```diff
@@ -323,6 +349,42 @@ UniValue getclaimsforname(const UniValue& params, bool fHelp)
...
+    } BOOST_SCOPE_EXIT_END
+
+    if (params.size() > 1)
+    {
+        int minconf = params[1].get_int();
+        while(--minconf > 0) {
+            indexes.push_back(chainActive.Tip());
+            CBlock block;
+            ReadBlockFromDisk(block, indexes.back(), Params().GetConsensus());
+            DisconnectBlock(block, state, indexes.back(), coins, claims);
```
BrannonKing: In this method we are actually rolling back the real blocks -- not just the caches. We hold the locks and disconnect some number of tips and then rely on our scope-exit handler to put everything back before we leave. I think it's a horrible approach, really, but didn't see a good way to do it otherwise in the current codebase. We lock the entire system for handling this method. That's so wrong. I think Roy's approach of storing all claimtrie (nodes that have ever existed) in LevelDB is much better -- the old nodes would still be there and you could just start a search at a certain time period -- we wouldn't ever have to lock; we could run numerous calls concurrently.

src/rpc/claimtrie.cpp
```diff
@@ -323,6 +349,42 @@ UniValue getclaimsforname(const UniValue& params, bool fHelp)
...
+            // assuming we don't need to resurrect the memory pool entries for this RPC call
+        }
+        coins.Flush(); // need flushes on every iteration?
+        claims.flush();
+    }
+
     claimsForNameType claimsForName = pclaimTrie->getClaimsForName(name);
```
BrannonKing: There was no equivalent of `getClaimsForName` in the cache object, and it was not obvious how to achieve this method's "support" outputs using the data in the cache.

src/claimtrie.cpp
lbrynaut: What is this supposed to do?  It doesn't make sense to me.

src/rpc/claimtrie.cpp
lbrynaut: All (current) claimtrie nodes are stored in leveldb.  Storing all changes over time (while nice for analysis) isn't actually that practical of a feature to have.  Eeven concurrent performance on lookups isn't needed for just about anything practical (other than analysis), where I'd argue a relational DB is just as well suited, if not better.  Horrible or not, all persistent state changes of a blockchain occur via connecting and disconnecting blocks.

Now is all this needed for this particular issue?  I'm not sure yet, haven't thought about it enough.

src/claimtrie.cpp
bvbfan: We need ability read cached claim, it makes sense when we use rollback feature.

src/claimtrie.cpp
```diff
@@ -2574,10 +2562,125 @@ uint256 CClaimTrieCache::getLeafHashForProof(const std::string& currentPosition,
...
+
+bool CClaimTrieCache::getInfoForName(const std::string& name, CClaimValue& claim) const
+{
+    if (dirty())
+        getMerkleHash();
+
+    CClaimTrieNode* current = &(base->root);
+    nodeCacheType::const_iterator cachedNode;
+
+    for (std::string::const_iterator itName = name.begin(); current; ++itName) {
```
BrannonKing: use `getNodeForName` here?

bvbfan: I've thinking about that, but getNodeForName fallback to base (CClaimTrie) when it's not found in cache.

bvbfan:
I plan to standardized JSON objects names, between function they have different names which is ugly and hard to parse in a script.
claimId
txId
txn
value
presentInTrie (in claim trie)
effectiveAmount
validAtHeight
and so on, first word starts with lower letter, second word, if it presents, starts with upper. Most help does not match returned JSON object, Any other suggestions?

BrannonKing:
> I plan to standardized JSON objects names...

We need to do that as a separate pull request. That's a breaking change, whereas adding additional outputs and taking new optional inputs are not breaking changes. I agree that the current mismatches between methods are silly. Do you know where each RPC method is used in the LBRY stack? We should locate each call and write up an epic story that includes updating each method that is necessary.

bvbfan:
> We need to do that as a separate pull request. That's a breaking change, whereas adding additional outputs and taking new optional inputs are not breaking changes. I agree that the current mismatches between methods are silly. 

I understand, you're right.

>Do you know where each RPC method is used in the LBRY stack? We should locate each call and write up an epic story that includes updating each method that is necessary.

lbry.go, i think, but @shyba knows better.

bvbfan:
Rebase to be ready for merge.

src/claimtrie.cpp
```diff
@@ -527,39 +527,26 @@ claimsForNameType CClaimTrie::getClaimsForName(const std::string& name) const
...
+        if (it->claimId == claimId && it->nValidAtHeight < nCurrentHeight) {
             effectiveAmount += it->nAmount;
-            claim_found = true;
+            for (std::vector<CSupportValue>::const_iterator it = claims.supports.begin(); it != claims.supports.end(); ++it) {
+                if (it->supportedClaimId == claimId && it->nValidAtHeight < nCurrentHeight) {
+                    effectiveAmount += it->nAmount;
+                    if (supports) supports->push_back(*it);
+                }
+            }
             break;
```
kaykurokawa: What are all these changes to getEffectiveAmoutnForClaim for? Seems like an unrelated change?

src/rpc/claimtrie.cpp
```diff
@@ -22,143 +24,177 @@ uint160 ParseClaimtrieId(const UniValue& v, const std::string& strName)
...
+    if (mapBlockIndex.count(blockHash) == 0)
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Block not found");
+
+    CBlockIndex* pblockIndex = mapBlockIndex[blockHash];
+    if (!chainActive.Contains(pblockIndex))
+        throw JSONRPCError(RPC_INTERNAL_ERROR, "Block not in main chain");
+
+    if (chainActive.Tip()->nHeight > (pblockIndex->nHeight + MAX_RPC_BLOCK_DECREMENTS))
+        throw JSONRPCError(RPC_INTERNAL_ERROR, "Block too deep to regenerate it");
+
```
kaykurokawa: Instead of "Block too deep to regenerate it" , how about just "Block is too deep"

src/rpc/claimtrie.cpp
```diff
@@ -22,143 +24,177 @@ uint160 ParseClaimtrieId(const UniValue& v, const std::string& strName)
...
             "Return all claims in the name trie.\n"
             "Arguments:\n"
-            "None\n"
+            "1. \"blockhash\"     (string, optional) the hash of the block\n"
+            "                                           which is the active\n"
+            "                                           in the chain. If\n"
+            "                                           none is given, \n"
+            "                                           the latest block\n"
+            "                                           will be used.\n"
             "Result: \n"
```
kaykurokawa: instead of "the hash of the block which is the active in the chain. If none is given, the latest block will be used", how about:

"get claims in the trie at the block specified by this block hash. If none is given, the latest active block will be used"

src/rpc/claimtrie.cpp
```diff
@@ -22,143 +24,177 @@ uint160 ParseClaimtrieId(const UniValue& v, const std::string& strName)
...
             "Return the entire name trie.\n"
             "Arguments:\n"
-            "None\n"
+            "1. \"blockhash\"   (string, optional) the hash of the block\n"
+            "                                         which is the active\n"
+            "                                         in the chain. If\n"
+            "                                         none is given, \n"
+            "                                         the latest block\n"
+            "                                         will be used.\n"
             "Result: \n"
```
kaykurokawa: how about:
"get the claim trie at the block specified by this block hash. If none is given, the latest active block will be used"

src/rpc/claimtrie.cpp
```diff
@@ -180,118 +216,112 @@ bool getValueForClaim(const COutPoint& out, std::string& sValue)
...
             "Arguments:\n"
-            "1. \"name\"             (string) the name to look up\n"
+            "1. \"name\"          (string) the name to look up\n"
+            "2. \"blockhash\"     (string, optional) the hash of the block\n"
+            "                                           which is the active\n"
+            "                                           in the chain. If\n"
+            "                                           none is given, \n"
+            "                                           the latest block\n"
+            "                                           will be used.\n"
             "Result: \n"
```
kaykurokawa: how about "get the value associated with the name at the block specified by this block hash. If none is given, the latest active block will be used"

src/rpc/claimtrie.cpp
```diff
@@ -180,118 +216,112 @@ bool getValueForClaim(const COutPoint& out, std::string& sValue)
...
             "Arguments: \n"
-            "1.  \"name\"       (string) the name for which to get claims and supports\n"
+            "1.  \"name\"         (string) the name for which to get claims and supports\n"
+            "2.  \"blockhash\"    (string, optional) the hash of the block\n"
+            "                                           which is the active\n"
+            "                                           in the chain. If\n"
+            "                                           none is given, \n"
+            "                                           the latest block\n"
+            "                                           will be used.\n"
             "Result:\n"
```
kaykurokawa: how about, "get claims for name at the block specified by this block hash. If none is given, the latest active block will be used"

kaykurokawa:
I added a commit for some unit tests for the new CClaimtrieCache functions in claimtriecache_tests.cpp, by no means extensive but I think it'd be good to have some coverage on those. 

https://github.com/lbryio/lbrycrd/pull/189/commits/0159a57765154ca2a09df4fc3845546eb3908a00

src/claimtrie.cpp
bvbfan: When I made a getEffectiveAmountForClaim in CClaimTrieCache notice it has 2 functions in here, so i just want to match them in the 2 classes. I prefer pointer to indicate it's optional or not. In plus now it has overloaded function that takes claimsForNameType if you performed getClaimsForName before to prevent double look-up. Otherwise we end-up with 4 functions
```
getEffectiveAmountForClaim(const std::string& name, uint160 claimId)
getEffectiveAmountForClaim(const claimsForNameType& claims, uint160 claimId)
getEffectiveAmountForClaimWithSupports(const std::string& name, uint160 claimId)
getEffectiveAmountForClaimWithSupports(const claimsForNameType& claims, uint160 claimId)
```

src/rpc/claimtrie.cpp
BrannonKing: getnameproof says "Block too deep to generate proof"

kaykurokawa:
Fixed formatting on the test I added.
Also noting that this will affect calls in https://github.com/lbryio/lbryumx for lbryumx/session.py , this should be the only place where this PR will affect. I will open issue there

BrannonKing:
How will libryumx be affected? The intent was that all new parameters here were optional.

bvbfan:
Update help strings.

src/rpc/claimtrie.cpp
BrannonKing: getnameproof says "Block too deep to generate proof"

BrannonKing:
I was just doing some testing by hand and seeing a problem:
1. start up a clean regtest daemon (after removing ~/.lbrycrd/regtest)
2. generate 200 blocks
3. claim a new name
4. generate 1 block
5. get the block hash for block 190
6. run getclaimsintrie passing in the hash for 190
7. the results should be empty but they aren't

kaykurokawa:
Noting here that RPC command getclaimbyid does not have a blockhash argument in this PR . 

I think it might be a bit involved to implement .. since we'll have to redo the getClaimById function in CClaimTrie ( it relies on leveldb and a cache, but the cache I think does not support decrementing multiple blocks at a time, just one block at at time) . We could probably open a new PR for this instead of  doing this here.

PR #191, opened 2018-08-20
refactor getValueForClaim

shyba:


kaykurokawa:
This LGTM with me

Adding here that the original conditional statement before the || has off by one error (outpoint.n could be equal to vout.size()) and than the following statement after the || could cause a segfault :

coin->vout.size() < itClaims->outPoint.n || coin->vout[itClaims->outPoint.n].IsNull()

kaykurokawa:
This has been merged https://github.com/lbryio/lbrycrd/commit/368a7389f4917534a6c9370d1ce5b104b8934a32

Issue #116, opened 2018-04-04
lbrycrdd allows broadcasting transactions containing invalid updates


jackrobison:
`broadcast` will send transactions containing claim update outputs where the claim id is invalid. Invalid claim id sizes as well as incorrect or non-existent claim ids all get through. Attempting to broadcast such transactions should return a transaction rejected error.

bvbfan:
Someone to work on?

kauffj:
@jackrobison if you're able include the specific examples of transactions you're able to broadcast but should not be able to, it'd make this ticket stronger.

bvbfan:
https://github.com/lbryio/lbrycrd/pull/192

lbrynaut:
Just noting that #192 does not solve the issue posted.

BrannonKing:
The `sendrawtransaction` method relays messages even if they aren't accepted into the local memory pool. This has been repaired in the upstream bitcoin, in mid 2016.

BrannonKing:
This is fixed now that we have the upstream master merged.

PR #192, opened 2018-08-20
Do not create transactions with incorrect claims

bvbfan:

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -426,6 +426,10 @@ UniValue claimname(const UniValue& params, bool fHelp)
         );
     string sName = params[0].get_str();
     string sValue = params[1].get_str();
+
+    if (!IsHex(sValue))
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "value must be hexadecimal");
+
     std::vector<unsigned char> vchName (sName.begin(), sName.end());
     std::vector<unsigned char> vchValue(ParseHex(sValue));
 
@@ -513,6 +517,10 @@ UniValue updateclaim(const UniValue& params, bool fHelp)
 
     std::vector<unsigned char> vchName;
     string sValue = params[1].get_str();
+
+    if (!IsHex(sValue))
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "value must be hexadecimal");
+
     std::vector<unsigned char> vchValue(ParseHex(sValue));
     CAmount nAmount = AmountFromValue(params[2]);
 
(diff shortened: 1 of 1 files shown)


bvbfan:
https://github.com/lbryio/lbrycrd/issues/116

src/wallet/rpcwallet.cpp
```diff
@@ -426,6 +427,14 @@ UniValue claimname(const UniValue& params, bool fHelp)
         );
     string sName = params[0].get_str();
     string sValue = params[1].get_str();
+
+    if (sName.empty())
+        throw JSONRPCError(RPC_WALLET_ERROR, "Claim name should not be empty");
+
```
kaykurokawa: while it is unintuitive, an empty claim name is actually valid (you can make claim for empty names when you use double quotes: "" in the RPC claimname command) so this check is not needed

src/wallet/rpcwallet.cpp
```diff
@@ -847,7 +864,23 @@ UniValue supportclaim(const UniValue& params, bool fHelp)
         );
 
     string sName = params[0].get_str();
+
+    if (sName.empty())
+        throw JSONRPCError(RPC_WALLET_ERROR, "Claim name should not be empty");
+
```
kaykurokawa: same here

src/wallet/rpcwallet.cpp
```diff
@@ -847,7 +864,23 @@ UniValue supportclaim(const UniValue& params, bool fHelp)
...
+        throw JSONRPCError(RPC_WALLET_ERROR, "Claim id should be exact 160 symbols");
+
+    size_t i = 0;
+    if (sClaimId[0] == 0 && tolower(sClaimId[1]) == 'x')
+        i = 2;
+
+    for (; i < sClaimId.size(); ++i)
+        if (!isxdigit(sClaimId[i]))
+            throw JSONRPCError(RPC_WALLET_ERROR, "Claim id should contains only hexdecimal numbers");
+
```
kaykurokawa: There is actually a function created to check claimId here: 
https://github.com/lbryio/lbrycrd/blob/master/src/rpc/claimtrie.cpp#L10

bvbfan:
rpc/claimtrie.cpp is a part of libbitcoin_server while rpcwallet of libbitcoin_wallet that's why i'm including only 2 checks from ParseClaimtrieId not compete one.

bvbfan:
Fix addressed issues.

bvbfan:
Mac build looks broken again.

BrannonKing:
Rebase off master again and the OSX build will be fixed.

bvbfan:
Done

PR #195, opened 2018-08-22
WIP: Refactor claim index [currently being investigated, dont merge]

shyba:


src/rpc/claimtrie.cpp
```diff
@@ -398,17 +398,16 @@ UniValue getclaimbyid(const UniValue& params, bool fHelp)
     LOCK(cs_main);
     uint160 claimId = ParseClaimtrieId(params[0], "Claim-id (parameter 1)");
     UniValue claim(UniValue::VOBJ);
-    std::string name;
     CClaimValue claimValue;
+    std::string name, sValue;
     pclaimTrie->getClaimById(claimId, name, claimValue);
-    if (claimValue.claimId == claimId)
+    if (pclaimTrie->getClaimById(claimId, name, claimValue) && getValueForClaim(claimValue.outPoint, sValue))
```
BrannonKing: `getClaimById` should not be called twice.

shyba: thanks, squashed and force pushed

src/rpc/claimtrie.cpp
```diff
@@ -398,17 +398,15 @@ UniValue getclaimbyid(const UniValue& params, bool fHelp)
...
     {
         std::vector<CSupportValue> supports;
         CAmount effectiveAmount = pclaimTrie->getEffectiveAmountForClaimWithSupports(
             name, claimValue.claimId, supports);
 
-        std::string sValue;
-        getValueForClaim(claimValue.outPoint, sValue);
+        claim.push_back(Pair("value", sValue));
         claim.push_back(Pair("name", name));
         claim.push_back(Pair("value", sValue));
```
lbrynaut: Duplicate of line 409 (remove either this or your addition)

src/test/claimtriebranching_tests.cpp
```diff
@@ -831,7 +831,7 @@ BOOST_AUTO_TEST_CASE(get_claim_by_id_test)
 {
     ClaimTrieChainFixture fixture;
     const std::string name = "test";
-    CMutableTransaction tx1 = fixture.MakeClaim(fixture.GetCoinbase(), name, "one", 1);
+    CMutableTransaction tx1 = fixture.MakeClaim(fixture.GetCoinbase(), name, "one", 2);
```
lbrynaut: Why is this changing?

src/test/claimtriebranching_tests.cpp
```diff
@@ -844,15 +844,28 @@ BOOST_AUTO_TEST_CASE(get_claim_by_id_test)
     fixture.MakeSupport(fixture.GetCoinbase(), tx1, name, 4);
     fixture.IncrementBlocks(1);
 
-    CMutableTransaction tx2 = fixture.MakeClaim(fixture.GetCoinbase(), name, "two", 1);
+    CMutableTransaction tx2 = fixture.MakeClaim(fixture.GetCoinbase(), name, "two", 2);
```
lbrynaut: Why is this changing?

src/test/claimtriebranching_tests.cpp
shyba: So the update can test the value stored at the index also gets updated from 2 to 1.

src/test/claimtriebranching_tests.cpp
shyba: typo, fixing. I meant to do this only to tx1

shyba:
squashed, fixed style, duplicated json field and typo on tests.

shyba:
Added a command to debug the issue:
```
./lbrycrd-cli help getmissingclaimsfromindex                                            
getmissingclaimsfromindex
Return the claims that arent on claims_by_id index.
Arguments:
None
Result:
[
  "name"           (string) the claimed name
  "claimId"        (string) the claim id
]
```
Please don't merge that commit. We can remove it once we have a definitive fix or the reason for that to be happening.

lbrynaut:
@shyba Thanks for bringing this issue up.  I looked into it a little bit and have a theory on what's going wrong, but will need some more time analyzing before testing a fix.  It's unrelated to this though, so no need to merge here.  If there's an existing issue related to this, please link it here.

shyba:
That one is actually a cosmetic byproduct from when I was investigating. It's some unit tests and a refactor. The only issue being solve in this PR is when the `value` field fails it won't return a result instead of returning an empty value. However that's also caused by the index issues.
If the refactor and the unit tests proves to be useful you can merge them, but otherwise we can close this PR. It's mostly investigation byproducts.

lbrynaut:
> If the refactor and the unit tests proves to be useful you can merge them, but otherwise we can close this PR. It's mostly investigation byproducts.

See #197.  If/when that's merged, please issue a PR to re-add the unit testing.

PR #202, opened 2018-09-04
WIP: Travis code format investigation (resolves #201)

### LINKED ISSUE #201: Travis builds sometimes cannot complete due to missing style scripts
Travis sometimes reports that it can't find the required clang formatting scripts.  This was seen originally on OS X (where it was arbitrarily disabled without investigating further for a fix) and it is now showing on GNU/Linux.

An example build failure shows as

```
./reproducible_build.sh: line 421: ./contrib/devtools/clang-format-diff.py: No such file or directory
+LINES_W_FORMAT_REQUIRED=0
```

An example log is here:

https://travis-ci.org/lbryio/lbrycrd/jobs/424021818

lbrynaut:


lbrynaut:
This does nothing, ignore it for now.  It's just allowing me to test things on travis until I find what's broken.

lbrynaut:
Interestingly, this did fix the issue in this case, as I suspected it had something to do with the way we're doing git clones and comparing cross-repo.  Not sure it's a permanent fix though (i.e. guaranteed to work in all cases).

lbrynaut:
Hrm, ok this commit appears to help it out better (ensuring we have the script before running it).  I think there are some states on travis that are more or less likely to fail, depending on what happened before it.

PR #209, opened 2018-10-03
implemented hard fork to include nonwinning claims in hash

BrannonKing:
It also uses Merkle trees for hashing children and claims post-fork. This significantly reduces the output for `getnameproof`.

Notable things to look at:
1. The way the tree is verified after it is loaded form disk (now using the Cache).
2. The place in the Cache's (Dis)ConnectBlock methods where we rehash the whole tree.
3. The use of the already-included MerkleHash and MerkleBranch methods.
4. The names in the getnameproof data and the Proof structs.

### DIFF
--- a/src/claimtrie.cpp
+++ b/src/claimtrie.cpp
@@ -1,10 +1,12 @@
 #include "claimtrie.h"
 #include "coins.h"
+#include "consensus/merkle.h"
 #include "hash.h"
 
+#include <algorithm>
 #include <boost/scoped_ptr.hpp>
 #include <iostream>
-#include <algorithm>
+#include <vector>
 
 std::vector<unsigned char> heightToVch(int n)
 {
@@ -21,7 +23,7 @@ std::vector<unsigned char> heightToVch(int n)
     return vchHeight;
 }
 
-uint256 getValueHash(COutPoint outPoint, int nHeightOfLastTakeover)
+uint256 getValueHash(const COutPoint& outPoint, int nHeightOfLastTakeover)
 {
     CHash256 txHasher;
     txHasher.Write(outPoint.hash.begin(), outPoint.hash.size());
@@ -134,7 +136,8 @@ void CClaimTrieNode::reorderClaims(supportMapEntryType& supports)
         }
     }
 
-    std::make_heap(claims.begin(), claims.end());
+    // max of effectiveAmount should be first and the rest should be sorted for repeatable hashing of them
+    std::sort(claims.rbegin(), claims.rend());
 }
 
 uint256 CClaimTrie::getMerkleHash()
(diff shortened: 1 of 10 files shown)


src/claimtrie.cpp
```diff
@@ -1237,20 +1212,80 @@ bool CClaimTrieCache::recursiveComputeMerkleHash(CClaimTrieNode* tnCurrent, std:
...
+bool CClaimTrieCache::recursiveComputeMerkleHashBinaryTree(CClaimTrieNode* tnCurrent, std::string sPos, bool forceCompute) const
+{
+    if (tnCurrent->empty()) {
+        cacheHashes[sPos] = _noChildrenOrClaims;
+        return true;
+    }
+
+    std::vector<uint256> childHashes;
+    for (nodeMapType::iterator it = tnCurrent->children.begin(); it != tnCurrent->children.end(); ++it) {
+        std::stringstream ss;
```
bvbfan: stringstream is slowest C++ string manipulation class, better not use it in time expensive functions. You can use string + char which is faster.

src/claimtrie.cpp
```diff
@@ -1237,20 +1212,80 @@ bool CClaimTrieCache::recursiveComputeMerkleHash(CClaimTrieNode* tnCurrent, std:
...
+        ss << it->first;
+        std::string sNextPos = sPos + ss.str();
+        if (forceCompute || dirtyHashes.count(sNextPos) != 0) {
```
bvbfan: count look-up through whole container constant, better use find and compare against end.

src/claimtrie.cpp
BrannonKing: Sorry, the stringstream was there in the existing method that I copied to start with. There are quite a few of those in the file, unfortunately. I completely agree with you that it's both slow and a horribly asinine way to convert a char to a string. I've been dragging my feet a little on completely refactoring the trie searches (aka, moving them all to a search method that all these other methods call); I've been waiting for the upstream merge to happen before doing some major refactoring on the trie. It would eliminate all stringstream calls.

src/claimtrie.cpp
BrannonKing: Again, both the stringstream and the count were in the previous method; I wasn't focused on those. If I do another round of changes on this story, I'll remove them though.

bvbfan: Yeah, i saw that but it's good time to fix it.

BrannonKing:
`CClaimTrieCache::insertClaimIntoTrie` needs to be updated to detect if there were any ordering changes.

src/claimtrie.cpp
```diff
@@ -1237,20 +1212,80 @@ bool CClaimTrieCache::recursiveComputeMerkleHash(CClaimTrieNode* tnCurrent, std:
     return true;
 }
```
kaykurokawa: I would like to see a better name for this function recursiveComputeMerkleHashBinaryTree or some comments/docstrings on what its supposed to do.

kaykurokawa:
We should clarify how the claims are ordered when hashed. 

It looks to me like the winning claim would always be the first in position of the merkle tree ( that is good because it is still important to be able to prove that a claim is the winning one). It also looks like the non winning claims are sorted from highest effective amount to lowest effective amount? I would make sure that this is the case (are claims ever stored out of order ?) .

BrannonKing:
We've had new requirements on this. It now needs to implement all necessary support for this: https://www.notion.so/lbry/Q-A-on-RPC-results-verification-44888d23efa3475a90eec997f9bf3103 . It also needs to return the claim_sequence field in all claim RPC calls.

PR #212, opened 2018-10-06
i add support to contact

SeboPG:


kaykurokawa:
Contact is not the same thing as support.
Support is provided on github.

PR #220, opened 2018-10-14
[WIP] Handle invalid bytes properly in normalization 

kaykurokawa:
This is a single commit 0f23d12 on top of #159 

#159 when normalizing has a way of handling invalid bytes (in claimtrie.h) as below where it tries to detect if the size of the normalized string shrank or not.

`return (normalized.size() >= name.size()) ? normalized : name;`

This coincidentally works because  invalid utf-8 strings after normalization are returned with a single byte truncated (I'm not sure if this is true for all invalid utf-8 bytes, I only checked a couple. A potential bad problem would be if an invalid utf-8 bytes after normalization are returned as valid utf-8 bytes with a size that is greater than it was before normalization). 

This process also has side effects because some valid utf8 strings that are normalized CAN shrink.  An example of this is unicode code point \x2126 which is the omega symbol. In utf-8, this is represented in bytes as \xE284A6. However, when normalized it is transformed to its equivalent unicode code point \x03A9, which is represented in bytes as \xCEA9. The size of the utf-8 has in this case, shrunk, but it is completely valid normalization.  ( Here is a useful tool to http://www.ltg.ed.ac.uk/~richard/utf-8.html look up utf-8 encoded characters)

Thus it is better to just detect invalid utf-8 strings (there was a method to do this through boost::locale) , and just return the invalid utf-8 string. 

Furthermore , I was testing the difference between the US_en locale, and the default C locale (also called POSIX locale) when lower casing. I noticed that the C locale did not have lower casing for non alphabetical letters, such as the above omega symbol ( https://en.wikipedia.org/wiki/Omega ). The US_en locale properly lower cased the omega symbol. 

However, it should also be noted that some users may want to use the non lower cased Omega symbol, for example to refer to ohms: https://en.wikipedia.org/wiki/Ohm . This is definitely not the only such symbol, where depending on the context of usage , lower casing may or may not make sense, so we probably want to consider this potential problem. 

I still need to add unit tests for this.


BrannonKing:
Moved by hand to #159

PR #229, opened 2018-10-25
Remove unused Python modules in linearize-data.py

gy741:
Hello,

In my opinion, No json, base64, httplib Python module are required.

Because it is not used.

What do you think of it?

Thanks.


BrannonKing:
I think that PRs need to address pre-existing issues and/or demonstrate significant work. This is an attempt to game the system; it's not appropriate.

PR #231, opened 2018-10-30
Serialize effective amount

bvbfan:


bvbfan:
It not the right approach.

PR #235, opened 2018-11-08
The main normalization branch that needs review pronto

kaykurokawa:
Some small commits on top of normalization PR #159.. The thread is getting quite lengthy so I thought I'd open a new PR...  @BrannonKing can quickly review this here and move these commits in there when he's ready. 

a0741b8 
Added tests that checks the normalization only , without making claims. Contains tests for invalid utf8, some tests for NFD normalization in certain languages, and casings on non english languages.

c26f2c9
Just moved and clarified a comment on the normalizeClaimName function

2f61fb2
Added an RPC call checknormalization so user can check what the normalization function does from RPC/CLI


BrannonKing:
This looks fine to me. I'll include it on the other branch shortly.

src/claimtrie.cpp
```diff
@@ -1290,33 +1286,34 @@ bool CClaimTrieCache::insertClaimIntoTrie(const std::string& name, CClaimValue c
...
         if (fCheckTakeover)
-            namesToCheckForTakeover.insert(name);
+            namesToCheckForTakeover.insert(newName);
     }
     return true;
 }
 
 bool CClaimTrieCache::removeClaimFromTrie(const std::string& name, const COutPoint& outPoint, CClaimValue& claim, bool fCheckTakeover) const
 {
+    const std::string newName = normalizeClaimName(name);
```
BrannonKing: I was trying to establish a boundary between the trie and the queues where we always normalize going into the trie. The same is true for the support map.

src/claimtrie.cpp
```diff
@@ -2114,8 +2108,16 @@ bool CClaimTrieCache::incrementBlock(insertUndoType& insertUndo, claimQueueRowTy
             }
             assert(found);
             const std::string name = normalizeClaimName(itEntry->first, forceNormalization);
+
+            // error triggered by claimtriecache_normalization:
+            // it will cause a double remove when rolling back -- something we don't handle correctly
+            // (same problem exists for the insertSupportUndo below)
+            //for (insertUndoType::reverse_iterator itInsertUndo = insertUndo.rbegin(); itInsertUndo != insertUndo.rend(); ++itInsertUndo)
+            //    if (itInsertUndo->name == itEntry->first && itInsertUndo->outPoint == itEntry->second.outPoint)
+            //        throw std::runtime_error("not expected");
```
BrannonKing: @bvbfan , this situation is for you to look at.

bvbfan:
We have 2 major issues in last attempt to keep track of original name, they aren't handled by test case, cause reordering nodes after normalization:
1. we loose ability to return same best claim
2. takeover height of can be away different
I've thinking about to minimize normalization effort by lower-casing on demand, note
```
std::string ut8_string = boost::locale::conv::to_utf<char>(name, "UTF-8", boost::locale::conv::stop);
if (utf8_string.empty())
    return name;
normalized = boost::locale::normalize(utf8_string, boost::locale::norm_nfd, utf8);

// lowering in case of real normalization
if (utf8_string != normalized)
    normalized = boost::locale::to_lower(normalized, utf8);
return normalized;
```
We can avoid lower-casing when we does not have *real* change.

src/claimtrie.cpp
```diff
@@ -1564,10 +1559,12 @@ bool CClaimTrieCache::removeClaimFromQueue(const std::string& name, const COutPo
     if (itQueueRow != claimQueueCache.end())
     {
         claimQueueRowType::iterator itQueue;
+        // compare normalized names for the UPDATE_OP; we allow you to change the case on an update after the fork
         for (itQueue = itQueueRow->second.begin(); itQueue != itQueueRow->second.end(); ++itQueue)
         {
-            if (name == itQueue->first && itQueue->second.outPoint == outPoint) {
-                break;
+            if (itQueue->second.outPoint == outPoint) {
+                if (normalizeClaimName(name) == normalizeClaimName(itQueue->first))
```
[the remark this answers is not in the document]
BrannonKing: I was expecting the outPoint to be reasonably unique, meaning that we would only call the normalization once or less on average.

src/claimtrie.cpp
```diff
@@ -1631,8 +1625,10 @@ void CClaimTrieCache::removeFromExpirationQueue(const std::string& name, const C
     {
         for (itQueue = itQueueRow->second.begin(); itQueue != itQueueRow->second.end(); ++itQueue)
         {
-            if (name == itQueue->name && outPoint == itQueue->outPoint)
-                break;
+            if (outPoint == itQueue->outPoint) {
+                if (normalizeClaimName(name) == normalizeClaimName(itQueue->name))
```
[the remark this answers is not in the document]
BrannonKing: Again here I was thinking that outPoint was unique.

BrannonKing:
> they aren't handled by test case, cause reordering nodes after normalization...
I did write a new test case to ensure that the nodes are reordered going forwards or backwards. (But I didn't do anything to the test to ensure that they were out of order at any point during the test.) 
>     we loose ability to return same best claim
I'm not sure what you're stating here exactly. We still return the best claim, but we now compare across a larger set of claims. It is true that previous best may come into competition with claims that it was not competing against before the fork.
>     takeover height can be away different
Only if the new competition causes it to change. Otherwise it should hold on to its previous value.

> I've thinking about to minimize normalization effort by lower-casing on demand, note
> std::string ut8_string = boost::locale::conv::to_utf<char>(name, "UTF-8", boost::locale::conv::stop);
> if (utf8_string.empty())
>     return name;
> normalized = boost::locale::normalize(utf8_string, boost::locale::norm_nfd, utf8);
> // lowering in case of real normalization
> if (utf8_string != normalized)
>     normalized = boost::locale::to_lower(normalized, utf8);
> return normalized;

This code doesn't make any sense to me. The goal was to make more names compete with each other. It's not just for normalization alone. We want to avoid having winners off by one letter casing, etc.

bvbfan:
>  It is true that previous best may come into competition with claims that it was not competing against before the fork.

I mean exactly that.

> Only if the new competition causes it to change.

It's true if you check that there isn't claims that after normalization will have same name and takeover height will override another.

> It's not just for normalization alone. We want to avoid having winners off by one letter casing, etc.

Yeah, in that case lower case makes sense,

src/claimtrie.cpp
```diff
@@ -1494,53 +1489,53 @@ queueNameType::iterator CClaimTrieCache::getQueueCacheNameRow(const std::string&
...
+        nameOutPointType entry(name, claim.outPoint);
         addToExpirationQueue(claim.nHeight + base->nExpirationTime, entry);
-        CClaimIndexElement element = {newName, claim};
+        CClaimIndexElement element = {name, claim};
         claimsToAdd.push_back(element);
-        return insertClaimIntoTrie(newName, claim, false);
+        return insertClaimIntoTrie(name, claim, false);
     }
-    return addClaimToQueues(newName, claim);
+    addClaimToQueues(name, claim);
```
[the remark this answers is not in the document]
BrannonKing: fixed

BrannonKing:
I just pushed in some minor changes and fixes. You will now see normalizedName in the RPC calls. (I haven't done a major RPC field name refactor yet.) I yanked all the code to return pending supports. I couldn't get the normalized name correct there, and the whole idea seemed wrong after further reflection. A support is not a support until the block increments. Same with claims. Please review it. Thanks.

src/claimtrie.cpp
BrannonKing: I think I understand how this is happening now. I have some ideas on how to address it.

bvbfan: Is this still happen? When I correct unit test it does not throw.

BrannonKing: It fails for me in latest with this test: claimtriebranching_tests/claimtriecache_normalization

bvbfan: Did you try that patch:
```
diff --git a/src/test/claimtriebranching_tests.cpp b/src/test/claimtriebranching_tests.cpp
index 48c5ce4f..6072947a 100644
--- a/src/test/claimtriebranching_tests.cpp
+++ b/src/test/claimtriebranching_tests.cpp
@@ -1212,7 +1212,6 @@ BOOST_AUTO_TEST_CASE(claimtriecache_normalization)
     BOOST_CHECK(name != name_upper);
 
     // Add another set of unicode claims that will collapse after the fork
-    CMutableTransaction tx1 = fixture.MakeClaim(fixture.GetCoinbase(), name, "amilie", 2);
     CMutableTransaction tx2 = fixture.MakeClaim(fixture.GetCoinbase(), name_upper, "amelie", 2);
     fixture.MakeClaim(fixture.GetCoinbase(), "amelie1", "amelie", 2);
     fixture.IncrementBlocks(1);
@@ -1231,6 +1230,7 @@ BOOST_AUTO_TEST_CASE(claimtriecache_normalization)
     fixture.setNormalizationForkHeight(1);
     int currentHeight = chainActive.Height();
 
+    CMutableTransaction tx1 = fixture.MakeClaim(fixture.GetCoinbase(), name, "amilie", 2);
     fixture.IncrementBlocks(1);
     // Ok normalization fix the name problem
     BOOST_CHECK(pclaimTrie->getInfoForName(name_normd, nval1));
@@ -1252,9 +1252,8 @@ BOOST_AUTO_TEST_CASE(claimtriecache_normalization)
     BOOST_CHECK(trieCache.spendClaim(name_upper, COutPoint(tx2.GetHash(), 0), currentHeight, amelieValidHeight));
 
     BOOST_CHECK(!pclaimTrie->getInfoForName(name, nval1));
-    BOOST_CHECK(trieCache.getInfoForName(name, nval1));
+    BOOST_CHECK(!trieCache.getInfoForName(name, nval1));
     BOOST_CHECK(trieCache.addClaim(name, COutPoint(tx1.GetHash(), 0), ClaimIdHash(tx1.GetHash(), 0), CAmount(2), currentHeight + 1));
-    BOOST_CHECK(trieCache.getInfoForName(name, nval1));
     insertUndoType insertUndo;
     claimQueueRowType expireUndo;
     insertUndoType insertSupportUndo;
```

src/claimtrie.cpp
```diff
@@ -1998,39 +1831,43 @@ bool CClaimTrieCache::undoSpendSupport(const std::string& name, const COutPoint&
     }
 }
 
-bool CClaimTrieCache::removeSupport(const std::string& name, const COutPoint& outPoint, int nHeight, int& nValidAtHeight, bool fCheckTakeover) const
+bool CClaimTrieCacheBase::removeSupport(const std::string& name, const COutPoint& outPoint, int nHeight, int& nValidAtHeight, bool fCheckTakeover) const
 {
     bool removed = false;
     CSupportValue support;
-    if (removeSupportFromQueue(name, outPoint, support))
+    nValidAtHeight = nHeight + getDelayForName(name);
```
BrannonKing: @lbrynaut , this is the line where I wanted you to verify that getDelayForName would return the same as in addClaim (aka, the same going forward as going backward).

src/rpc/claimtrie.cpp
```diff
@@ -142,11 +143,15 @@ UniValue getclaimsintrie(const UniValue& params, bool fHelp)
...
+            std::string targetName;
+            CClaimValue targetClaim;
+            if (pclaimTrie->getClaimById(itClaims->claimId, targetName, targetClaim))
+                claim.push_back(Pair("name", targetName));
             claims.push_back(claim);
         }
 
         UniValue node(UniValue::VOBJ);
-        node.push_back(Pair("name", it->first));
+        node.push_back(Pair("normalizedName", it->first));
```
BrannonKing: This needs to be 'normalized_name'. After looking through the lbry codebase today, I saw how much they really want PEP8 names.

src/rpc/claimtrie.cpp
```diff
@@ -465,13 +487,16 @@ UniValue getclaimbyid(const UniValue& params, bool fHelp)
...
+        CClaimTrieCache cache(pclaimTrie);
         std::vector<CSupportValue> supports;
-        CAmount effectiveAmount = pclaimTrie->getEffectiveAmountForClaim(name, claimValue.claimId, &supports);
+        CAmount effectiveAmount = cache.getEffectiveAmountForClaim(name, claimValue.claimId, &supports);
 
         std::string sValue;
         CCoinsViewCache coins(pcoinsTip);
         getValueForClaim(coins, claimValue.outPoint, sValue);
         claim.push_back(Pair("name", name));
+        if (cache.shouldNormalize())
```
BrannonKing: I'm thinking to change this to always return the normalized name, not just after the fork. That we people can adjust to it sooner.

configure.ac
```diff
@@ -742,12 +749,20 @@ if test x$use_pkgconfig = xyes; then
     AC_MSG_ERROR(pkg-config not found.)
   fi
 
+  if test "${ICU_PREFIX}" != "auto"; then
```
lbrynaut: ICU detection is still broken.  What this is doing is preferring the system ICU over the built dependency ICU version, which is not what we want.  Our build must require --with-icu and prefer that path.  I'll update with something later and you can see if it works for you.

PR #238, opened 2018-11-23
Increase default dbcache size

bvbfan:


src/rpc/claimtrie.cpp
```diff
@@ -35,12 +33,43 @@ static CBlockIndex* BlockHashIndex(const uint256& blockHash)
...
+
+        if (ShutdownRequested())
+            throw JSONRPCError(RPC_INTERNAL_ERROR, "Shutdown requested");
+
+        CValidationState state;
+        if (!DisconnectBlock(block, state, index, coinsCache, trieCache))
+            throw JSONRPCError(RPC_INTERNAL_ERROR, strprintf("Failed to disconnect %s", block.ToString()));
+
+        if (state.IsError())
+            return false;
```
BrannonKing: Why not throw here as well? and make the method void return?

src/txdb.h
```diff
@@ -20,7 +20,7 @@ struct CDiskTxPos;
 class uint256;
 
 //! -dbcache default (MiB)
-static const int64_t nDefaultDbCache = 100;
+static const int64_t nDefaultDbCache = 700;
```
BrannonKing: I'm worried about this default on Windows. It's currently running in 32bit and using 1.2GB of RAM. This will bump it up against the 2GB barrier on Windows.

src/rpc/claimtrie.cpp
```diff
@@ -35,12 +33,43 @@ static CBlockIndex* BlockHashIndex(const uint256& blockHash)
     if (!chainActive.Contains(pblockIndex))
         throw JSONRPCError(RPC_INTERNAL_ERROR, "Block not in main chain");
 
+    const size_t currentMemoryUsage = pcoinsTip->DynamicMemoryUsage();
+    const int MAX_RPC_BLOCK_DECREMENTS = nCoinCacheUsage / currentMemoryUsage;
```
BrannonKing: I like the idea of limiting decrements on RAM. However, I don't think this line accomplishes that. You can't make an assumption that each block uses the same amount of coin cache. I think this check of MAX decrements could probably go away. We just use available memory then give up.

bvbfan: It *relaxing* check if it want some deeper block that is *easy* knowing that rollback will fail.

src/txdb.h
bvbfan: So 450 or 500 as Kay suggest?

BrannonKing: What about `sizeof(void*) >= 8 ? 700 : 400`.

src/rpc/claimtrie.cpp
BrannonKing: We don't know that 100 blocks back will fail. It's dependent upon the number of TXOs in each block.

src/txdb.h
bvbfan: That's good.

src/rpc/claimtrie.cpp
bvbfan: Yep but most likely will fail, i think we need some value that will prevent rollback to take time for nothing. nCoinCacheUsage / currentMemoryUsage * 1.5 or something ?

src/rpc/claimtrie.cpp
bvbfan: Looks like state is totally unused in DisconnectBlock i'll remove that check entirely.

src/rpc/claimtrie.cpp
kaykurokawa: I agree with Brannon that this approach

'MAX_RPC_BLOCK_DECREMENTS = nCoinCacheUsage / currentMemoryUsage' is not correct. 

because current memory usage is not predictive of what memory usage a roll back will cost. 

I'm in favor of keeping MAX_RPC_BLOCK_DECREMENTS to a high constant. I think the previous max of 50 is fine, or we could divide nCoinCacheUsage by 20 mb (in my tests I saw that a 1mb block used about 10 mb of cache). This is  in case someone accidentally sets a very high roll back, we don't have to waste time or resource trying to roll back, we will know immediately that the roll back is too high.

src/rpc/claimtrie.cpp
```diff
@@ -35,12 +37,48 @@ static CBlockIndex* BlockHashIndex(const uint256& blockHash)
...
-
     return pblockIndex;
 }
 
+void RollBackTo(const CBlockIndex* targetIndex, CCoinsViewCache& coinsCache, CClaimTrieCache& trieCache)
+{
+    AssertLockHeld(cs_main);
+    const size_t txCount = pcoinsTip->GetCacheSize();
+    const CBlockIndex* activeIndex = chainActive.Tip();
+    const size_t currentMemoryUsage = pcoinsTip->DynamicMemoryUsage();
```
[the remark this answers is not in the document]
bvbfan: On empty block we can fallback to @kaykurokawa suggest
>  think the previous max of 50 is fine, or we could divide nCoinCacheUsage by 20 mb

What you think?

PR #241, opened 2018-12-04
Replace flattenTrie by callback

bvbfan:


kaykurokawa:
I don't really understand the "InterruptionPoint" part of the code. Can you explain a little bit how it works and what its' for? Is it based on some design pattern I can read about somewhere?

bvbfan:
> I think I understand but it seems like we don't need this feature because we iterate through everything? 

Yes, we iterate on all trie right now, but we can check shutdown signal in.

> Is this needed for some future use case?

Yep, this approach combines features of std::for_each and std::find_if in one. It gives flexibility and performance, since we are at ~480k height iteration on all trie is expensive.

PR #253, opened 2019-02-08
fixed blocks coming in too fast on testnet

BrannonKing:
When we get a delayed block, we thereafter knock the difficulty so low that it takes multiple (up to a 100) blocks to get back up to sufficient difficulty. It doesn't make sense for us to do this as we already adjust difficulty every block.

### DIFF
--- a/src/pow.cpp
+++ b/src/pow.cpp
@@ -18,7 +18,7 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
     if (pindexLast == NULL)
         return nProofOfWorkLimit;
 
-    if (params.fPowAllowMinDifficultyBlocks && pindexLast->nHeight >= 277299)
+    if (params.fPowAllowMinDifficultyBlocks && pindexLast->nHeight >= 277299 && pindexLast->nHeight < 1100000)
     {
         // Special difficulty rule for testnet:
         // If the new block's timestamp is twice the target block time
@@ -29,12 +29,15 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
         if (pblock->GetBlockTime() > pindexLast->GetBlockTime() + params.nPowTargetSpacing*2){
             return nProofOfWorkLimit;
         }
+        // Well, actually, it wasn't ever implemented properly in LBRYcrd.
+        // It doesn't work without the missing "else" statement found in upstream bitcoin.
+        // And that statement doesn't work correctly with our DifficultyAdjustmentInterval == 1.
+        // Hence, we are killing it at block 1100000.
     }
 
     // Go back the full period unless it's the first retarget after genesis.
-    int blockstogoback = params.DifficultyAdjustmentInterval()-1;
-    if ((pindexLast->nHeight+1) != params.DifficultyAdjustmentInterval())
-        blockstogoback = params.DifficultyAdjustmentInterval();
+    int blockstogoback = params.DifficultyAdjustmentInterval();
+    blockstogoback = std::min(blockstogoback, pindexLast->nHeight);
 
     int nHeightFirst = pindexLast->nHeight - blockstogoback;
     assert(nHeightFirst >= 0);
--- a/src/pow.cpp
+++ b/src/pow.cpp
@@ -18,7 +18,8 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
     if (pindexLast == NULL)
         return nProofOfWorkLimit;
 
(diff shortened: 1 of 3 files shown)


src/pow.cpp
```diff
@@ -18,7 +18,7 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
     if (pindexLast == NULL)
         return nProofOfWorkLimit;
 
-    if (params.fPowAllowMinDifficultyBlocks && pindexLast->nHeight >= 277299)
+    if (params.fPowAllowMinDifficultyBlocks && pindexLast->nHeight >= 277299 && pindexLast->nHeight < 1100000)
```
lbrynaut: Move both fork heights to chainparams

src/pow.cpp
```diff
@@ -29,12 +29,15 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
         if (pblock->GetBlockTime() > pindexLast->GetBlockTime() + params.nPowTargetSpacing*2){
             return nProofOfWorkLimit;
         }
+        // Well, actually, it wasn't ever implemented properly in LBRYcrd.
+        // It doesn't work without the missing "else" statement found in upstream bitcoin.
+        // And that statement doesn't work correctly with our DifficultyAdjustmentInterval == 1.
+        // Hence, we are killing it at block 1100000.
     }
```
lbrynaut: I agree it hasn't worked properly, but I think returning the difficulty back to something sane is better than letting testnet potentially get stuck.  The entire point of the testnet is to make testing easier/possible, so the extra headache if ever stuck is just an added barrier.  Doing nothing arguably allows easy testing (and mining) to continue as well.

src/pow.cpp
```diff
@@ -29,12 +29,15 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
...
 
     // Go back the full period unless it's the first retarget after genesis.
-    int blockstogoback = params.DifficultyAdjustmentInterval()-1;
-    if ((pindexLast->nHeight+1) != params.DifficultyAdjustmentInterval())
-        blockstogoback = params.DifficultyAdjustmentInterval();
+    int blockstogoback = params.DifficultyAdjustmentInterval();
+    blockstogoback = std::min(blockstogoback, pindexLast->nHeight);
```
lbrynaut: What's the point here?  I'd actually think that all of this could probably be removed since we re-target every block and our full period is constant.

src/pow.cpp
BrannonKing: They only apply to testnet (because that's the only one with fPowAllowMinDifficultyBlocks enabled). Still think it's worth it?

src/pow.cpp
BrannonKing: It's allowed to move by an eighth (12.5%) every block. You can travel a long way in half an hour -- something most people could tolerate. Something sane might be the median of the past 50 blocks; it could get complicated. 

Judging form the numbers I was seeing, though, there are miners going in and out of this often. I don't know why that is. I know that our testnet servers are on shared machines; maybe their resources fluctuate significantly. Or maybe the addition of a third person is sporadic but not uncommon.

src/pow.cpp
BrannonKing: Yes, it's only effective the very first time its hit (on the block right after genesis). We could special-case that block, but that would be making an assumption that DifficultyAdjustmentInterval returns 1. It does return 1 with our current settings, but if we're going to go down that road we might want to eliminate the parameters. And after working on this, I think that adjusting difficulty every block is unwise. It's too noisy, too prone to simple CPU fluctuations.

src/pow.cpp
lbrynaut: Yes, all forks should be documented/handled there.  The params in the other networks could be anything (i.e. 0 or 1 or whatever) since it should never apply.

src/pow.cpp
lbrynaut: Unwise or not, that's how the network operates.  I don't think it's any wiser to fork for a modification here, especially since mainnet has been running more or less sanely (i.e. nothing is broken at the moment).  And it's even less sane to do different retarget periods for testnet than for mainnet.

BrannonKing: Yes, the cap on the block height should make testnet and mainnet the same again. And I agree that a fork to make us not readjust difficulty every block is likely not worth it.

src/pow.cpp
BrannonKing: Fixed.

src/pow.cpp
BrannonKing: I left my `min` call in as I think it's much more clear than what was there before, but feel free to revisit this in the upstream merge.

bvbfan:
I'm OK with changes, let's @lbrynaut makes final review.

PR #258, opened 2019-03-05
WIP: Updated rebase 0.17

lbrynaut:

PR #259, opened 2019-03-06
WIP: Updated rebase 0.17 deps

lbrynaut:


bvbfan:
Can we keep BOOST_CHECK_EQUAL(a, b) over BOOST_CHECK(a == b), the second one do not show the 2 values when it fails.

lbrynaut:
> Can we keep BOOST_CHECK_EQUAL(a, b) over BOOST_CHECK(a == b), the second one do not show the 2 values when it fails.

Yes.  BTW, this is still WIP, not expecting it'll merge soon.  It's also an alternate/experimental take on #258

PR #266, opened 2019-05-02
Rebase lbry on to Bitcoin 0.17.

lbrynaut:
This contains significant rebase / merge / testing work by Naut
<lbrynaut@protonmail.com>, Anthony Fieroni <bvbfan@abv.bg> and Brannon
King <countprimes@gmail.com>.


src/util.cpp
```diff
@@ -73,69 +73,18 @@
...
-const char * const BITCOIN_CONF_FILENAME = "bitcoin.conf";
-const char * const BITCOIN_PID_FILENAME = "bitcoind.pid";
+const char * const BITCOIN_CONF_FILENAME = "lbrycrd.conf";
+const char * const BITCOIN_PID_FILENAME = "lbrycrd.pid";
 
 ArgsManager gArgs;
 
 CTranslationInterface translationInterface;
 
-/** Init OpenSSL library multithreading support */
```
BrannonKing: Why is this removed?

src/wallet/wallet.cpp
```diff
@@ -1690,18 +1699,19 @@ void CWalletTx::GetAmounts(std::list<COutputEntry>& listReceived,
 
         // In either case, we need to get the destination address
         CTxDestination address;
+        const CScript& scriptPubKey = StripClaimScriptPrefix(txout.scriptPubKey);
```
BrannonKing: I don't think we want this. We don't want to count our claims as spendable change. The old wallet didn't have it.

src/wallet/wallet.cpp
```diff
@@ -1403,11 +1410,13 @@ bool CWallet::IsChange(const CTxOut& txout) const
     if (::IsMine(*this, txout.scriptPubKey))
     {
         CTxDestination address;
-        if (!ExtractDestination(txout.scriptPubKey, address))
+        const CScript& scriptPubKey = StripClaimScriptPrefix(txout.scriptPubKey);
```
BrannonKing: This seems wrong; we don't want to count our held claims as available change.

src/test/denialofservice_tests.cpp
```diff
@@ -146,7 +146,14 @@ BOOST_AUTO_TEST_CASE(stale_tip_peer_management)
         BOOST_CHECK(node->fDisconnect == false);
     }
 
-    SetMockTime(GetTime() + 3*consensusParams.nPowTargetSpacing + 1);
+    // STALE_CHECK_INTERVAL is used in PeerLogicValidation::CheckForStaleTipAndEvictPeers
+    // as minimal time to check tip stale i.e. 10 minutes
+    // we use maximum value of STALE_CHECK_INTERVAL and nPowTargetSpacing
+    // NOTE: STALE_CHECK_INTERVAL is static that why we use raw value 10 * 60, sync may need in future
+
+    auto time = std::max({static_cast<int64_t>(10 * 60), 3 * consensusParams.nPowTargetSpacing});
```
BrannonKing: braces?

src/util.cpp
lbrynaut: OpenSSL was removed from the core code.  Every project should remove it.

src/wallet/wallet.cpp
lbrynaut: Not sure what you mean.  https://github.com/lbryio/lbrycrd/blob/legacy_master/src/wallet/wallet.cpp#L1152

src/wallet/wallet.cpp
lbrynaut: https://github.com/lbryio/lbrycrd/blob/legacy_master/src/wallet/wallet.cpp#L1004

src/wallet/wallet.cpp
BrannonKing: My bad. I was looking in the wrong spot.

src/test/denialofservice_tests.cpp
lbrynaut: Could be wrong, but I think this was a test @bvbfan worked on.  In any case, std::max can take an initializer list, so there's really no difference (functionally) if they're removed other than style preference.

PR #274, opened 2019-05-08
restored the current "depends" and friends

BrannonKing:
Goals with this PR:

1. We should use the dependency versions published by bitcoin. Exceptions include boost::locale fixes, the clang compiler for OSX (to use newer C++), and any security fixes we're aware of (none).
2. We should use "depends" for our official Linux, Windows, and OSX builds so that they all have the same dependencies.
3. The name "reproducible" needs to go away, as that's a misnomer. However, we should provide docker images for a nearly-reproducible experience (and use them on the build server).
4. Compiling the traditional way (./configure ...) should work; our scripts shouldn't be required.
5. Ensure that the Travis caching doesn't have unintended side effects.

For reviewing this, first check which files match bitcoin v17. That will eliminate the need for reviewing the majority of the files.


bvbfan:
For 3: I like how it's done by Dockefile, it's reproducible to me. We can build images as well. Alternatively we can use a package manager like Conan.
For 4: Should we go for CMake as better build system over autotools?

BrannonKing:
I like the idea of supporting Conan, and of changing out our use of build.lbry.io/lbrycrd in favor of Conan. I think that's a separate story, though. 

I definitely prefer CMake over Autotools. However, I don't think we can make that switch without the upstream bitcoin project also changing. It would be too heavy of maintenance on our part every time we merge from upstream. See https://github.com/bitcoin/bitcoin/issues/14118

BrannonKing:
This solution at present requires ccache to be installed for any of the packaging scripts. I would prefer it to be optional. I think that could be introduced as a separate PR.

BrannonKing:
In answer to the question of what bug does this fix: I had it happen to me again yesterday. The build for the 12.4.1 release failed on the OSX build with an incompatible binary error. I had to clear the TravisCI cache to make it work. This implies that ccache is confusing cross-platform files in the current code. That is addressed by the changes in this PR.

PR #275, opened 2019-05-09
allow rest/block/height.json

BrannonKing:
This is for Flipside crypto. It allows them to use our SSL rest endpoint (rest.lbry.tech) for their query analysis.

### DIFF
--- a/src/rest.cpp
+++ b/src/rest.cpp
@@ -205,14 +205,26 @@ static bool rest_block(HTTPRequest* req,
     const RetFormat rf = ParseDataFormat(hashStr, strURIPart);
 
     uint256 hash;
-    if (!ParseHashStr(hashStr, hash))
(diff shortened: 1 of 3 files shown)


src/rest.cpp
```diff
@@ -205,14 +205,26 @@ static bool rest_block(HTTPRequest* req,
     const RetFormat rf = ParseDataFormat(hashStr, strURIPart);
 
     uint256 hash;
-    if (!ParseHashStr(hashStr, hash))
-        return RESTERR(req, HTTP_BAD_REQUEST, "Invalid hash: " + hashStr);
-
+    long int blockHeight = 0;
+    if (hashStr != "tip") {
+        blockHeight = hashStr.size() < 12 ? std::strtol(hashStr.c_str(), nullptr, 10) : 0;
```
bvbfan: Some info or comment that we expect signed integer.

BrannonKing: Fixed

PR #276, opened 2019-05-11
Memory improvements, claimtrie revamp

### LINKED ISSUE #106: Reimplement ClaimTrie as Patricia Trie

### LINKED ISSUE #108: Separate CClaimTrie into trie implementation and ClaimTrie / Write base template trie class
The CClaimTrie class, which implements the Claim Trie, has a tight integration between the trie data structure and the logic involved in claims. 

It would be nice if we had a base template trie class which the CClaimTrie class either inherits from, or used as a member variable (CClaimTrie would either way use a trie class where the node are CClaimTrieNodes). This will improve code readability and allow us to write unit tests exclusively for the trie class to verify its correctness instead of relying on integration tests. 

Also,  it will allow easier implementation of https://github.com/lbryio/lbrycrd/issues/106, as we should just be able to switch out the trie class while keeping the logic for claims intact.

### LINKED ISSUE #133: Something goes wrong when saving claimtrie nodes to disk and loading it
Discovered in https://github.com/lbryio/lbrycrd/pull/110 

When loading in claimtrie nodes using function InsertFromDisk() https://github.com/lbryio/lbrycrd/blob/master/src/claimtrie.cpp#L1072 , we discovered that for some reason the "name" variable contains an extra bogus character at the end.

However, I suspect that there is not always a bogus character attached to the "name" variable, thus this function will completely fail. 

I've tested the serialization classes to see if there is any problem loading/saving string classes and there is no problem, so its likely a problem originating from when we save nodes, particularly when the function markNodeDirty() gets called https://github.com/lbryio/lbrycrd/blob/master/src/claimtrie.cpp#L828 which marks nodes to be saved on disk.

### LINKED ISSUE #138: nCurrentHeight in ClaimTrie and ClaimTrieCache needs to be renamed
nCurrentHeight, a member variable in ClaimTrie and ClaimTrieCache, is not the actual current height. 
It is the current height + 1 (i.e, if there is a total of two blocks including the genesis, nCurrentHeight is 2 where as the standard definition of current height would be 1) , or perhaps it can be called as the current height "to be worked on". 

nNextHeight might be a better name.

### LINKED ISSUE #145: CClaimTrieCache should be renamed
The naming of CClaimTrieCache in claimtrie.h/.cpp has caused a lot of confusion. 
It is not a Cache for CClaimTrie, it is a cache of pending changes to CClaimTrie upon block increment/decrement.

### LINKED ISSUE #158: Properly manage the allocation of CClaimTrieNode
CClaimTrieNode (nodes on a claimtrie) is currently allocated/de-allocated through the new/delete operators which has caused some problems in the past: https://github.com/lbryio/lbrycrd/pull/94

Using shared pointers would be a good improvement over this, so this could be the first step on this issue. Furthermore, we need to restructure the code so that we have a fixed size memory pool that limits the number of CClaimTrieNode objects that can be loaded into memory, while the rest is kept on disk (at this point the code will create CClaimTrieNode objects until it runs out of memory)   

This PR was started to solve this problem https://github.com/lbryio/lbrycrd/pull/100 using the PIMPL idiom but was left unfinished. I think it has the right idea, but it ran into a bug that we could not resolve. Further work on this should at least involve looking over this PR to see if it is a good approach, and suggest alternative approach if not. 

 ### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

BrannonKing:
This PR addresses the following issues:

Fixes #106 (Reimplement claimtrie without so many empty nodes, save RAM)
Fixes #108 (Separate data structure and data operations)
Fixes #133 (Trie serialization failure)
Fixes #138 (Rename nCurrentHeight)
Fixes #145 (Rename CClaimTrieCache)
Fixes #158 (Properly delete CClaimTrieNode)

It also includes these enhancements:
1. BLOCK_HASH is no longer needlessly stored in the claimtrie DB. I had a theory that this was masking places that we should have been comparing the claimTrieHash, which theory proved correct.
2. An issue with takeover height computation was detected and (temporarily) worked around.
3. cachedTakeoverHeights was removed; we now use the takeoverHeight field on the nodes in the cache.
4. The getQueue* functions now re-use code.
5. effectiveAmount is updated when reading from disk (we should now ensure that it's not being needlessly computed elsewhere).

Although the commits have been squashed, much of this work was done by @bvbfan .

We experimented with keeping empty nodes truly empty -- allocating claim space for them as needed. This saved less than 10MB of RAM and complicated the code a bit; we ditched it. We also experimented with a similar approach where empty node hashes were kept in a separate map. This also was too complicated to get right.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1503,12 +1503,6 @@ bool AppInitMain()
                     break;
                 }
 
-                if (!pclaimTrie->ReadFromDisk(true))
-                {
-                    strLoadError = _("Error loading the claim trie from disk");
-                    break;
-                }
-
                 // At this point we're either in reindex or we've loaded a useful
                 // block tree into mapBlockIndex!
 
@@ -1539,6 +1533,13 @@ bool AppInitMain()
                         break;
                     }
                     assert(chainActive.Tip() != nullptr);
+                }
+
+                CClaimTrieCache trieCache(pclaimTrie);
+                if (!trieCache.ReadFromDisk(chainActive.Tip()))
+                {
+                    strLoadError = _("Error loading the claim trie from disk");
+                    break;
                 }
 
                 if (!fReset) {
@@ -1702,7 +1703,7 @@ bool AppInitMain()
     LogPrintf("nBestHeight = %d\n", chain_active_height);
 
     const Consensus::Params& consensusParams = Params().GetConsensus();
-    pclaimTrie->setExpirationTime(consensusParams.GetExpirationTime(chain_active_height));
+    CClaimTrieCache(pclaimTrie).setExpirationTime(consensusParams.GetExpirationTime(chain_active_height));
 
     if (gArgs.GetBoolArg("-listenonion", DEFAULT_LISTEN_ONION))
         StartTorControl();
--- a/src/prefixtrie.cpp
+++ b/src/prefixtrie.cpp
@@ -1,0 +1,492 @@
+
+#include "prefixtrie.h"
+#include "claimtrie.h"
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+CPrefixTrie<TKey, TData>::Iterator<IsConst>::Iterator(const TKey& name, const std::shared_ptr<Node>& node) noexcept : name(name), node(node)
+{
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+template <bool C>
+typename CPrefixTrie<TKey, TData>::template Iterator<IsConst>& CPrefixTrie<TKey, TData>::Iterator<IsConst>::operator=(const CPrefixTrie<TKey, TData>::Iterator<C>& o) noexcept
+{
+    name = o.name;
+    node = o.node;
+    stack.clear();
+    stack.reserve(o.stack.size());
+    for (auto& i : o.stack)
+        stack.push_back(Bookmark{i.name, i.parent, i.it, i.end});
+    return *this;
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+bool CPrefixTrie<TKey, TData>::Iterator<IsConst>::hasNext() const
+{
+    auto shared = node.lock();
+    if (!shared) return false;
+    if (!shared->children.empty()) return true;
+    for (auto it = stack.rbegin(); it != stack.rend(); ++it) {
+        auto mark = *it; // copy
+        if (++mark.it != mark.end)
+            return true;
+    }
+    return false;
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+typename CPrefixTrie<TKey, TData>::template Iterator<IsConst>& CPrefixTrie<TKey, TData>::Iterator<IsConst>::operator++()
+{
+    auto shared = node.lock();
+    assert(shared);
+    // going in pre-order (NLR). See https://en.wikipedia.org/wiki/Tree_traversal
+    // if there are any children we have to go there first
+    if (!shared->children.empty()) {
+        auto& children = shared->children;
+        auto it = children.begin();
+        stack.emplace_back(Bookmark{name, shared, it, children.end()});
+        auto& postfix = it->first;
+        name.insert(name.end(), postfix.begin(), postfix.end());
+        node = it->second;
+        return *this;
+    }
+
+    // move to next sibling:
+    while (!stack.empty()) {
+        auto& back = stack.back();
+        if (++back.it != back.end) {
+            name = back.name;
+            auto& postfix = back.it->first;
+            name.insert(name.end(), postfix.begin(), postfix.end());
+            node = back.it->second;
+            return *this;
+        }
+        stack.pop_back();
+    }
+
+    // must be at the end:
+    node.reset();
+    name = TKey();
+    return *this;
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+typename CPrefixTrie<TKey, TData>::template Iterator<IsConst> CPrefixTrie<TKey, TData>::Iterator<IsConst>::operator++(int x)
+{
+    auto ret = *this;
+    ++(*this);
+    return ret;
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+CPrefixTrie<TKey, TData>::Iterator<IsConst>::operator bool() const
+{
+    return !node.expired();
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+bool CPrefixTrie<TKey, TData>::Iterator<IsConst>::operator==(const Iterator& o) const
+{
+    return node.lock() == o.node.lock();
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+bool CPrefixTrie<TKey, TData>::Iterator<IsConst>::operator!=(const Iterator& o) const
+{
+    return !(*this == o);
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+typename CPrefixTrie<TKey, TData>::template Iterator<IsConst>::reference CPrefixTrie<TKey, TData>::Iterator<IsConst>::operator*() const
+{
+    assert(!node.expired());
+    return TPair(name, *(node.lock()->data));
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+typename CPrefixTrie<TKey, TData>::template Iterator<IsConst>::pointer CPrefixTrie<TKey, TData>::Iterator<IsConst>::operator->() const
+{
+    assert(!node.expired());
+    return node.lock()->data.get();
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
+const TKey& CPrefixTrie<TKey, TData>::Iterator<IsConst>::key() const
+{
+    return name;
+}
+
+template <typename TKey, typename TData>
+template <bool IsConst>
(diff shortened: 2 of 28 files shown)


src/logging.h
```diff
@@ -17,7 +17,7 @@
 #include <vector>
 
 static const bool DEFAULT_LOGTIMEMICROS = false;
-static const bool DEFAULT_LOGIPS        = false;
+static const bool DEFAULT_LOGIPS        = true;
```
tiger5226: @lbrynaut Is this meant to be here?

BrannonKing: I'm glad you brought this up. I'm the one that changed the default and checked it in. I found it frustrating during our last hard fork that none of the log files had IP addresses in them. I suppose this is there because some people may be uncomfortable exposing their internal IP addresses. I think that concern is generally a thing of the past, though. I'm certainly open to arguments the other way. I think we should enable IP logging on all instances that we manage.

src/prefixtrie.cpp
```diff
@@ -0,0 +1,486 @@
...
+    return const_iterator{TKey(), std::shared_ptr<Node>{}};
+}
+
+using Key = std::string;
+using Data = CClaimTrieData;
+using Trie = CPrefixTrie<Key, Data>;
+using iterator = Trie::iterator;
+using const_iterator = Trie::const_iterator;
+
+template class CPrefixTrie<Key, Data>;
```
BrannonKing: This and the lines below it are a c++ nuance. You have to instantiate a specific type of a template or you need the full definition in the header file. As a native C# programmer, I prefer the latter. @bvbfan prefers the former (as it is here).

src/prefixtrie.cpp
```diff
@@ -0,0 +1,486 @@
...
+}
+
+template <typename TKey, typename TData>
+CPrefixTrie<TKey, TData>::CPrefixTrie() : size(0), root(std::make_shared<Node>())
+{
+    root->data = std::make_shared<TData>();
+}
+
+template <typename TKey, typename TData>
+template <typename TDataUni>
```
BrannonKing: TDataUni allows us to have both lvalue and rvalue data inputs on this method. The alternative is to have two overloads of this method.

src/prefixtrie.cpp
```diff
@@ -0,0 +1,486 @@
...
+typename CPrefixTrie<TKey, TData>::iterator CPrefixTrie<TKey, TData>::insert(CPrefixTrie<TKey, TData>::iterator& it, const TKey& key, TDataUni&& data)
+{
+    auto shared = it.node.lock();
+    assert(shared);
+    auto copy = it;
+    if (!key.empty()) {
+        auto name = it.key();
+        name.insert(name.end(), key.begin(), key.end());
+        auto& node = insert(key, shared);
+        copy = iterator{name, node};
```
BrannonKing: Not a true forward iterator -- that would require the full stack. This is a children-only iterator. It's true of all the places we create the iterator here. We are specifically choosing to not initialize them fully, and we're using the weak_ptr workaround for the place where we need the parent function but don't have it.

src/logging.h
tiger5226: Agreed. If we find it useful we should enable it, but maybe disable it by default for privacy.

src/validation.h
```diff
@@ -200,7 +200,7 @@ static const unsigned int MIN_BLOCKS_TO_KEEP = 288;
 /** Minimum blocks required to signal NODE_NETWORK_LIMITED */
 static const unsigned int NODE_NETWORK_LIMITED_MIN_BLOCKS = 288;
 
-static const signed int DEFAULT_CHECKBLOCKS = 6;
+static const signed int DEFAULT_CHECKBLOCKS = 24;
```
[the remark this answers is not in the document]
tiger5226: this is probably a very low priority attack vector, but 6 blocks with lbrycrd is not a lot to overwrite with malware. Most exchanges have min confirmations at 200 for LBC. Validating the top 200 blocks is probably a good default for right now. Again the likelihood that enough people catch the LBC malware that does this is probably near zero...just a thought though after researching the default value.

src/test/claimtriecache_tests.cpp
```diff
@@ -219,17 +209,18 @@ BOOST_AUTO_TEST_CASE(basic_insertion_info_test)
     int height = 0;
     int validHeight = 0;
     CClaimValue claimVal(claimOutPoint, claimId, amount, height, validHeight);
-    ctc.insertClaimIntoTrie("test", claimVal);
+    ctc.insertClaimIntoTrie("test", claimVal, true);
```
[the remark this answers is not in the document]
BrannonKing: claimtrie.h::insertClaimIntoTrie utilizes a default parameter value even though it's a virtual method. This is non-standard c++, and whether or not inheritors use that default is undefined. I had intended to remove it, but I apparently only made it part way down that road.

src/prefixtrie.cpp
bvbfan: In some project is used *.impl file ext (not *.cpp) and gets include where it's needed, i'm ok with that too. But all in header makes things to look redundant and hard to read.

src/test/claimtriecache_tests.cpp
```diff
@@ -257,84 +252,78 @@ BOOST_AUTO_TEST_CASE(recursive_prune_test)
...
-
-        void visit(const std::string& name, const CClaimTrieNode* node)
-        {
-            if (--count <= 0)
-                throw CRecursionInterruptionException(false);
+    }
+    BOOST_CHECK_EQUAL(count, 2);
+
+    count = 0;
+    // we cannot use range for on nodes whitout data
```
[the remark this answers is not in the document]
bvbfan: The comment is old, it should be removed at all, we have data on all nodes so it's save to be used range-based for loop.

src/prefixtrie.cpp
bvbfan: That's by design, iterator is forward through children, i mean this is not an exception. Thus we don't have ability to obtain parent, that's nodes stand for. We can go for *full* iterator but i'm afraid we will fill iterator stack for nothing i.e. we will not use it.

src/validation.h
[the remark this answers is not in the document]
lbrynaut: Possibly unrelated to this PR, but that's a good point. Bitcoin core did backtrack this from 288 to 6, but it's likely not appropriate for us.  Reverting to 288 is probably more sane.

src/logging.h
lbrynaut: I'd leave it disabled as well.

src/init.cpp
```diff
@@ -1541,6 +1535,13 @@ bool AppInitMain()
                     assert(chainActive.Tip() != nullptr);
                 }
 
+                CClaimTrieCache trieCache(pclaimTrie);
+                if (!trieCache.ReadFromDisk(chainActive.Tip()))
```
BrannonKing: I still feel like this method fits better on the base class.

src/init.cpp
```diff
@@ -1702,7 +1703,7 @@ bool AppInitMain()
     LogPrintf("nBestHeight = %d\n", chain_active_height);
 
     const Consensus::Params& consensusParams = Params().GetConsensus();
-    pclaimTrie->setExpirationTime(consensusParams.GetExpirationTime(chain_active_height));
+    CClaimTrieCache(pclaimTrie).setExpirationTime(consensusParams.GetExpirationTime(chain_active_height));
```
BrannonKing: This method also fits better on the base class.

src/validation.h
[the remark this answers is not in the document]
BrannonKing: 24 is an hour -- same as bitcoin. Why would we need more than that?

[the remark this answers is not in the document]
tiger5226: I don't think time is the intention here. The intention is to confirm enough blocks such that it is statistically painful to overwrite like with a big reorg for a double spend attack. 6 blocks of bitcoin is a standard for confirmation not meant to be a time equivalence. Admittedly, it is nice that it evenly breaks down to an hour, however arbitrary, but should not be the correlation between cryptocurrencies. However, I could be misinterpreting it.

src/rpc/claimtrie.cpp
```diff
@@ -103,112 +129,27 @@ static bool getValueForOutPoint(const CCoinsViewCache& coinsCache, const COutPoi
...
-    return ret;
+    auto count = params.size();
+    return count == required || count == required + optional;
 }
 
-static UniValue getclaimtrie(const JSONRPCRequest& request)
+static UniValue getclaimsintrie(const JSONRPCRequest& request)
 {
-    if (request.fHelp || request.params.size() > 1)
+    if (request.fHelp || !validParams(request.params, 0, 1))
```
tiger5226: not that I am well informed on whats going on here, but @BrannonKing wouldn't it be better to handle this sort of thing in the processing of the request rather than modifying so many lines of code? Seems like a DRY violation.

PR #284, opened 2019-06-17
allow local (non-pool) mining with ccminer and sgminer

BrannonKing:
support coinbasetxn capability in getblocktemplate


lbrynaut:
Tested well, locally.

PR #290, opened 2019-07-02
Add cmake build system

bvbfan:


bvbfan:
We can use cmake as reproducible build system, as well as rolling one. By default it's done to perform a monolith (static) build of all depends, on other hand it can be used to perform dynamic build against system libraries. The flexibility is that when depend is not found in system libs it's compiled as static one, so we have exactly what you need. On the other hand i have an idea to manage a branch that includes diffs by file against concrete bitcoin version so you can reproduce exactly the version you want. Since that is not very interesting in previous version, it'll be pretty much to newer one, since you can easy try to apply all patches on top of newer bitcoin version to generate a lbrycrd one. When can tag changes in diffs like tags in bitcoin which will make process easier, despite that menage and review diffs is not that easy task, it should be done by manually testing but furthermore i think this will make future painless.

BrannonKing:
I did some testing on this today. Here are issues that I see with it:
1. The boost build doesn't correctly locate the ICU libs. This is a must-have. "Performing configuration checks" must report that it found the ICU. I played with this for a little while in trying to make it more like the boost.mk that we use presently. However, I was unable to make it work.
2. Do we really need a separate b2 call for headers?
3. The Boost and ICU and other builds don't account for the RELEASE vs DEBUG build flags in CMake.
4. Use Boost 1.69 and ICU 63.2.
5. It was unable to locate my local boost libraries.
6. If use a local boost library, we don't need to use ICU.
7. We don't need program_options boost component, and the test component is named "test".
8. I think lbrycrd's --enable-cxx is deprecated.
9. It would be nice to use cmake vars instead of using "x86_64", etc. directly.

bvbfan:
> Do we really need a separate b2 call for headers?

Yes.

> If use a local boost library, we don't need to use ICU.

I don't get it. We need ICU when we build boost locally, we don't need if we use system wide one (because it's prebuild with icu in).

> and the test component is named "test"

Component is called unit_test_framework, i used in local build too.

> It would be nice to use cmake vars instead of using "x86_64", etc. directly.

Good catch, i'll correct it.

Can you show the output, i've test it in all variants, boost locate icu in my test cases.

BrannonKing:
I tested this with Clion today. It doesn't work. However, I have found a new and better way to make Clion work with the autotools stuff. Essentially, you use compiledb to build the compile_commands.json file from the make file. This works better than that older CMakeLists.txt file I've been using. With that in mind, I'm thinking to get rid of the CMakeLists.txt file. This newer CMake approach allows a mix of local and downloaded dependencies. That is an improvement over the existing build system, where it's either all or nothing on local dependencies (or it's some custom build work). At the same time, I don't want to confuse our users by having multiple ways to build the project. Another advantage of CMake is that it keeps the compiled files separated from the sources. However, the present approach to using Externals doesn't quite achieve this goal.

cmake/CPM.cmake
```diff
@@ -0,0 +1,210 @@
...
+    endif()
+    return()
+  endif()
+endif()
+
+set(CPM_VERSION ${CURRENT_CPM_VERSION} CACHE INTERNAL "")
+set(CPM_DIRECTORY ${CMAKE_CURRENT_LIST_DIR} CACHE INTERNAL "")
+set(CPM_PACKAGES "" CACHE INTERNAL "")
+
+option(CPM_USE_LOCAL_PACKAGES "Use locally installed packages (find_package)" OFF)
```
BrannonKing: Would it be better to default this to ON? I kind of like the idea of using local dependencies by default.

bvbfan:
What exactly fails?

BrannonKing:
Clion must be using the files passed to the add_executable/library Cmake commands. Those aren't called as part of the ExternalProject stuff. Like I said, though, this other approach with compiledb actually works much better than what I had.

CMakeLists.txt
```diff
@@ -1,56 +1,186 @@
...
+
+    if(NOT ${BerkeleyDB_VERSION} VERSION_LESS "5.0")
+        set(OPTIONS "${OPTIONS} --with-incompatible-bdb")
+    endif()
+
+    if(BerkeleyDB_ADDED)
+        ExternalProject_Add(BerkeleyDB
+            PREFIX bdb
+            SOURCE_DIR ${BerkeleyDB_SOURCE_DIR}
+            PATCH_COMMAND patch -Nd ${BerkeleyDB_SOURCE_DIR}/dbinc -i ${CMAKE_CURRENT_SOURCE_DIR}/depends/patches/db/atomic.patch || true
```
BrannonKing: The "depends" system uses `sed` instead of `patch` to achieve this. It doesn't need the extra file. The `sed` approach seems superior to me.

bvbfan:
I just test with KDevelop, it works, open project, import CMakeFiles.txt and then run build.

BrannonKing:
No, I did the version number very specifically: I want to download the 69 if I'm going to download one. If I'm not going to download one, then version 64+ will be fine.

PR #291, opened 2019-07-02
Code refactor

bvbfan:

### DIFF
--- a/src/claimtrie.cpp
+++ b/src/claimtrie.cpp
@@ -42,59 +42,37 @@ uint256 getValueHash(const COutPoint& outPoint, int nHeightOfLastTakeover)
 }
 
 template <typename T>
+bool equals(const T& lhs, const T& rhs)
+{
+    return lhs == rhs;
+}
+
+template <typename T>
 bool equals(const T& value, const COutPoint& outPoint)
 {
     return value.outPoint == outPoint;
 }
 
-template <typename T>
-bool equals(const T& value, const std::string& name, const COutPoint& outPoint)
-{
-    return value.name == name && value.outPoint == outPoint;
-}
-
 template <typename K, typename V>
-bool equals(const std::pair<K, V>& pair, const std::string& name, const COutPoint& outPoint)
+bool equals(const std::pair<K, V>& pair, const CNameOutPointType& point)
 {
-    return pair.first == name && pair.second.outPoint == outPoint;
+    return pair.first == point.name && pair.second.outPoint == point.outPoint;
 }
 
-template <typename T>
-auto findOutPoint(T& cont, const COutPoint& outPoint) -> decltype(cont.begin())
+template <typename T, typename C>
+auto findOutPoint(T& cont, const C& point) -> decltype(cont.begin())
 {
     using type = typename T::value_type;
     static_assert(std::is_same<typename std::remove_const<T>::type, std::vector<type>>::value, "T should be a vector type");
-    return std::find_if(cont.begin(), cont.end(), [&outPoint](const type& val) {
-        return equals(val, outPoint);
+    return std::find_if(cont.begin(), cont.end(), [&point](const type& val) {
+        return equals(val, point);
     });
 }
 
-template <typename T>
-auto findOutPoint(T& cont, const std::string& name, const COutPoint& outPoint) -> decltype(cont.begin())
+template <typename T, typename C>
+bool eraseOutPoint(std::vector<T>& cont, const C& point, T* value = nullptr)
 {
-    using type = typename T::value_type;
-    static_assert(std::is_same<typename std::remove_const<T>::type, std::vector<type>>::value, "T should be a vector type");
-    return std::find_if(cont.begin(), cont.end(), [&name, &outPoint](const type& val) {
-        return equals(val, name, outPoint);
-    });
-}
-
-template <typename T>
-bool eraseOutPoint(std::vector<T>& cont, const COutPoint& outPoint, T* value = nullptr)
-{
-    auto it = findOutPoint(cont, outPoint);
-    if (it == cont.end())
-        return false;
-    if (value)
-        std::swap(*it, *value);
-    cont.erase(it);
-    return true;
-}
-
-template <typename T>
-bool eraseOutPoint(std::vector<T>& cont, const std::string& name, const COutPoint& outPoint, T* value = nullptr)
-{
-    auto it = findOutPoint(cont, name, outPoint);
+    auto it = findOutPoint(cont, point);
     if (it == cont.end())
         return false;
     if (value)
@@ -269,7 +247,7 @@ bool CClaimTrieCacheBase::haveInQueue(const std::string& name, const COutPoint& 
         if (itNameRow != nameRow->second.end()) {
             nValidAtHeight = itNameRow->nHeight;
             if (auto row = getQueueCacheRow<T>(nValidAtHeight)) {
-                auto iRow = findOutPoint(row->second, name, outPoint);
+                auto iRow = findOutPoint(row->second, CNameOutPointType{name, outPoint});
                 if (iRow != row->second.end()) {
                     if (iRow->second.nValidAtHeight != nValidAtHeight)
                         LogPrintf("%s: An inconsistency was found in the support queue. Please report this to the developers:\nDifferent nValidAtHeight between named queue and height queue\n: name: %s, txid: %s, nOut: %d, nValidAtHeight in named queue: %d, nValidAtHeight in height queue: %d current height: %d\n", __func__, name, outPoint.hash.GetHex(), outPoint.n, nValidAtHeight, iRow->second.nValidAtHeight, nNextHeight);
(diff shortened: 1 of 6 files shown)


bvbfan:
This code refactoring aims to templatize all doublets algorithms addClaim/SupportToQueque, removeClaim/SupportFromQueue etc. to make code base well readable and better maintainable. We have issues with partial bug fixes due to same algorithm on more than one place, so this PR tends to remove than code duplication.

src/claimtrie.cpp
```diff
@@ -41,6 +41,68 @@ uint256 getValueHash(const COutPoint& outPoint, int nHeightOfLastTakeover)
     return result;
 }
 
+template <typename T>
+bool equals(const T& value, const COutPoint& outPoint)
+{
+    return value.outPoint == outPoint;
+}
+
+template <typename T>
```
BrannonKing: Can we make CSupportValue and CClaimValue have the same base class that has the name and outpoint so that we don't have to use templated methods to work with them?

src/claimtrie.cpp
```diff
@@ -41,6 +41,68 @@ uint256 getValueHash(const COutPoint& outPoint, int nHeightOfLastTakeover)
...
+{
+    using type = typename T::value_type;
+    static_assert(std::is_same<typename std::remove_const<T>::type, std::vector<type>>::value, "T should be a vector type");
+    return std::find_if(cont.begin(), cont.end(), [&outPoint](const type& val) {
+        return equals(val, outPoint);
+    });
+}
+
+template <typename T>
+auto findOutPoint(T& cont, const std::string& name, const COutPoint& outPoint) -> decltype(cont.begin())
```
BrannonKing: `-> decltype` is c++14, no?

src/claimtrie.cpp
```diff
@@ -118,47 +173,82 @@ typename Container::value_type* getQueue(CDBWrapper& db, uint8_t dbkey, const Ke
...
-typename claimQueueType::value_type* CClaimTrieCacheBase::getQueueCacheRow(int nHeight, bool createIfNotExists)
+template <>
+std::pair<const int, std::vector<queueEntryType<CClaimValue>>>* CClaimTrieCacheBase::getQueueCacheRow(int nHeight, bool createIfNotExists)
 {
     return getQueue(*(base->db), CLAIM_QUEUE_ROW, nHeight, claimQueueCache, createIfNotExists);
 }
 
-typename queueNameType::value_type* CClaimTrieCacheBase::getQueueCacheNameRow(const std::string& name, bool createIfNotExists)
+template <>
+std::pair<const int, std::vector<queueEntryType<CSupportValue>>>* CClaimTrieCacheBase::getQueueCacheRow(int nHeight, bool createIfNotExists)
```
BrannonKing: how does this method name imply support?

src/claimtrie.cpp
```diff
@@ -118,47 +173,82 @@ typename Container::value_type* getQueue(CDBWrapper& db, uint8_t dbkey, const Ke
...
+}
+
+template <>
+typename expirationQueueType::value_type* CClaimTrieCacheBase::getExpirationQueueCacheRow<CSupportValue>(int nHeight, bool createIfNotExists)
 {
     return getQueue(*(base->db), SUPPORT_EXP_QUEUE_ROW, nHeight, supportExpirationQueueCache, createIfNotExists);
 }
 
+template <typename T>
+typename expirationQueueType::value_type* CClaimTrieCacheBase::getExpirationQueueCacheRow(int, bool)
```
BrannonKing: Why is this stubbed method here?

src/claimtrie.cpp
```diff
@@ -312,57 +374,72 @@ void completeHash(uint256& partialHash, const std::string& key, std::size_t to)
...
+    };
+
+    iCbType<iterator> process = [&failed, &process, &verify](iterator& it) {
+        if (it->hash != recursiveMerkleHash(it, process, verify)) {
+            failed = it.key();
+            throw CRecursiveBreak();
+        }
+    };
+
+    try {
```
BrannonKing: A recursive break is not an exceptional condition. This should use return values instead. try/catch should be extremely rare and the catch should never be the expected scenario.

src/claimtrie.cpp
bvbfan: We can but dynamic dispatch is runtime / slower vs static one. If we go for polymorphic approach we'll face same penalty as well as runtime check to determine actual type of parameter.

src/claimtrie.cpp
bvbfan: No, https://en.cppreference.com/w/cpp/language/decltype

src/claimtrie.cpp
bvbfan:
```
std::pair<const int, std::vector<queueEntryType<CSupportValue>>>*
                                                -------------
````

File: src/claimtrie.cpp (hunk 4 above)
Comment (bvbfan) [reply src/claimtrie.cpp]: The idea of meta programming is determining the type in compile time, i.e. using a function with unwanted type can fail compile time. The *stubs* are functions that will fail at compile time i.e. incorrect usage, supportedType<T> will fail compilation for all type but claim and support.

File: src/claimtrie.cpp (hunk 5 above)
Comment (bvbfan) [reply src/claimtrie.cpp]: It is https://github.com/lbryio/lbrycrd/pull/291/commits/35a987bf2b7bf7b4d4f5b08297c5a45a4d467492#diff-f8129adb47c1b30eac7d2795b17883edR427
That's easy way to stop recursion, it does not affect other exceptions.

File: src/claimtrie.cpp (hunk 3 above)
Comment (BrannonKing) [reply src/claimtrie.cpp]: Differentiating methods based on return value is not a good practice. Most computer languages don't even allow this; they don't include the return value type in the method footprint.

File: src/claimtrie.cpp (hunk 5 above)
Comment (BrannonKing) [reply src/claimtrie.cpp]: How is it easier than using a return value?

File: src/claimtrie.cpp (hunk 1 above)
Comment (BrannonKing) [reply src/claimtrie.cpp]: Taking a base class as a parameter doesn't automatically make a virtual function, no?

File: src/claimtrie.cpp (hunk 3 above)
Comment (bvbfan) [reply src/claimtrie.cpp]: It's perfectly valid in C++ and it's one of must have features to meta programming.

File: src/claimtrie.cpp (hunk 5 above)
Comment (bvbfan) [reply src/claimtrie.cpp]: It breaks the stack without unwinding, i.e. it does not pop all stack pushes and the stack is unwind when catch appears. https://en.wikipedia.org/wiki/Call_stack#Unwinding

File: src/claimtrie.cpp (hunk 1 above)
Comment (bvbfan) [reply src/claimtrie.cpp]: Yes, there is not automatically virtual functions, but you get the base class at some point of implementation you should what exactly is this type, claim or support without virtual dispatch you can't do it.

Comment (bvbfan) [reply src/claimtrie.cpp]: The C approach is same type for both (claim and support) with field in (enum in most cases) that describe the actual type.

File: src/claimtrieforks.cpp [hunk 6]
```diff
@@ -153,27 +129,25 @@ bool CClaimTrieCacheNormalizationFork::normalizeAllNamesInTrieIfNecessary(insert
             if (support.nHeight + base->nExpirationTime <= nNextHeight)
                 continue;
 
-            CSupportValue removed;
-            assert(removeSupportFromMap(it.key(), support.outPoint, removed, false));
-            expireSupportUndo.emplace_back(it.key(), removed);
+            assert(removeSupportFromMap(it.key(), support.outPoint, support, false));
```
BrannonKing: This has a bug in that it modifies the item we're about to insert. That other variable was there on purpose.

src/claimtrieforks.cpp
```diff
@@ -153,27 +129,25 @@ bool CClaimTrieCacheNormalizationFork::normalizeAllNamesInTrieIfNecessary(insert
...
 
         for (auto& claim : it->claims) {
             if (claim.nHeight + base->nExpirationTime <= nNextHeight)
                 continue;
 
-            CClaimValue removed;
-            assert(removeClaimFromTrie(it.key(), claim.outPoint, removed, false));
-            removeUndo.emplace_back(it.key(), removed);
-            assert(insertClaimIntoTrie(normalized, claim, false));
+            assert(removeClaimFromTrie(it.key(), claim.outPoint, claim, false));
```
BrannonKing: Same here: the other variable is necessary.

src/claimtrie.cpp
```diff
@@ -312,57 +352,72 @@ void completeHash(uint256& partialHash, const std::string& key, std::size_t to)
...
 {
     std::vector<uint8_t> vchToHash;
-
     const auto pos = it.key().size();
-    auto children = it.children();
-    for (auto& child : children) {
-        if (!recursiveCheckConsistency(child, minimumHeight, failed))
-            return false;
+    for (auto& child : it.children()) {
+        process(child);
```
BrannonKing: "process" is pretty vague. Do we have a more specific name? And this is just to allow a few lines of code reuse between validation and hash computation?

src/claimtrieforks.cpp
bvbfan: Actually it isn't, because we know that we remove same support then read it under new name, no changes in value.

src/claimtrie.cpp
bvbfan: Suggest better naming, i'll update it.

src/claimtrieforks.cpp
bvbfan: But better to iterate through copy rather than reference.

src/claimtrie.cpp
```diff
@@ -469,8 +524,7 @@ bool CClaimTrieCacheBase::ReadFromDisk(const CBlockIndex* tip)
 {
     LogPrintf("Loading the claim trie from disk...\n");
 
-    nNextHeight = tip ? tip->nHeight + 1 : 0;
-    base->nNextHeight = nNextHeight;
+    base->nNextHeight = nNextHeight = tip ? tip->nHeight + 1 : 0;
```
BrannonKing: Missing an `==`

bvbfan: There is no `==` here, it assigns tip height to nNextHeight then to base one.

PR #296, opened 2019-07-10
Fix expiration fork usage

### LINKED ISSUE #287: expiration fork isn't set in miner.cpp
The fork expiration isn't set in miner.cpp. This causes a failure when testing mining on regtest (when mining past the expiration fork -- use the fuzz test). This is because the mined block doesn't have the same expiration time that the block validation uses.

I'd like to see us remove the expiration time state from claimtrie and claimtriecache. It should check the expiration height computation in overloaded methods like we do for normalization. This won't be any worse performance because we do the check in validation.cpp on every increment/decrement anyhow.

bvbfan:

Fixes #287

### DIFF
--- a/src/test/claimtriebranching_tests.cpp
+++ b/src/test/claimtriebranching_tests.cpp
@@ -498,7 +498,7 @@ BOOST_AUTO_TEST_CASE(triehash_fuzzer_test)
     }
 
     if (blocks == 1000 && claimsPerBlock == 100)
-        BOOST_CHECK_EQUAL(fixture.getMerkleHash().GetHex(), "c5df693adbe6b386a5d1e494e83a8f5745f32f6917a32b4f429e7c20b47171dd");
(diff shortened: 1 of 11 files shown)


src/test/claimtriebranching_tests.cpp
```diff
@@ -1723,7 +1723,7 @@ BOOST_AUTO_TEST_CASE(normalization_does_not_kill_sort_order)
 BOOST_AUTO_TEST_CASE(normalization_does_not_kill_expirations)
 {
     ClaimTrieChainFixture fixture;
-    fixture.setExpirationTime(3);
+    fixture.setExpirationForkHeight(800, 3, 800);
```
BrannonKing: I don't think this is right. Aren't those 800s supposed to be 140, the current height?

bvbfan: In ClaimTrieChainFixture we don't change the consensus, so setExpirationTime was supposed to changed only expiration time, since we don't have such functionality it should be
```
fixture.setExpirationForkHeight(consensus.nExtendedClaimExpirationForkHeight, 3, consensus.nExtendedClaimExpirationTime)
```

PR #299, opened 2019-07-28
Fix lbrycrd-cli command typo

AlessandroSpallina:
From version 0.16 the use of argument "getinfo" with lbrycrd-cli return the error code -32601; 
as suggested in the resulting output, use instead "-getinfo" in order to get info as the old format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ the "--help" flag (e.g. `lbrycrdd --help`). Examples:
 
 Run `./lbrycrdd -server -daemon` to start lbrycrdd in the background.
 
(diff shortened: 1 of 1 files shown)


tzarebczan:
Hey, thanks for the PR! I'm pretty sure this command is obsolete and replaced with other get* commands for specific sets of information - see https://lbry.tech/api/blockchain

AlessandroSpallina:
Yes, you are right the "getinfo" is deprecated, but "-getinfo" still works by calling some "get*" and returns like the old style output

PR #303, opened 2019-08-14
[WIP] Unify and extend rpc methods to bid and sequence

bvbfan:
A bunch of updated rpc keys can lead to breaking changes in users, note it uses camelCase name notation, it can be changed as well. Notable changes:
nomalized_name -> normalizedName
txid -> txId
nHeight -> height
nValidAtHeight (valid at height) -> validAtHeight
nAmount -> amount
nEffectiveAmount (effective amount) -> effectiveAmount
total names -> totalNames
total claims -> totalClaims
total value -> totalValue
Most values in getclaimsfortx

### DIFF
--- a/src/rpc/claimrpchelp.h
+++ b/src/rpc/claimrpchelp.h
@@ -100,7 +100,7 @@ S3("        ", T_AMOUNT, "             (numeric) the amount of the support") \
 S1("    ]") \
 S3("    ", T_LASTTAKEOVERHEIGHT, "     (numeric) the last height at which ownership of the name changed") \
 S3("    ", T_BID, "                    (numeric) lower value means a higher bid rate, ordered by effective amount") \
-S3("    ", T_SEQUENCE, "               (numeric) lower value means latest in sequence, ordered by height of insertion")
+S3("    ", T_SEQUENCE, "               (numeric) lower value means an older one in sequence, ordered by height of insertion")
 
 #define PROOF_OUTPUT \
 S3("    ", T_NODES, ": [               (array of object, pre-fork) full nodes\n" \
@@ -194,7 +194,7 @@ S3("            ", T_VALIDATHEIGHT, "  (numeric) the height at which the support
 S3("            ", T_AMOUNT, "         (numeric) the amount of the support")
 S1("        ]")
 S3("        ", T_BID, "                (numeric) lower value means a higher bid rate, ordered by effective amount")
-S3("        ", T_SEQUENCE, "           (numeric) lower value means latest in sequence, ordered by height of insertion")
+S3("        ", T_SEQUENCE, "           (numeric) lower value means an older one in sequence, ordered by height of insertion")
 S1("    ]")
 S3("    ", T_LASTTAKEOVERHEIGHT, "     (numeric) the last height at which ownership of the name changed")
 S3("    ", T_SUPPORTSWITHOUTCLAIM, ": [")
--- a/src/rpc/claimrpchelp.h
+++ b/src/rpc/claimrpchelp.h
@@ -1,0 +1,290 @@
+
+#include <string>
+
+// always keep defines T_ + value in upper case
+#define T_NORMALIZEDNAME                "normalizedName"
+#define T_BLOCKHASH                     "blockhash"
+#define T_CLAIMS                        "claims"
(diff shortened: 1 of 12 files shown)


src/rpc/claimrpchelp.h
```diff
@@ -0,0 +1,341 @@
...
+
+#define NAME_TEXT      "                    (string) the name to look up"
+#define BLOCKHASH_TEXT "               (string, optional) get claims in the trie\n" \
+"                                                  at the block specified\n" \
+"                                                  by this block hash.\n" \
+"                                                  If none is given,\n" \
+"                                                  the latest active\n" \
+"                                                  block will be used."
+
+#define CLAIM_OUTPUT    \
```
tiger5226: These are less readable than before. I actually used these quite often.

src/rpc/claimtrie.cpp
```diff
@@ -102,91 +107,194 @@ std::string escapeNonUtf8(const std::string& name)
...
+    if (element.claim.claimId == claimId) {
+        name = element.name;
+        if (claim)
+            *claim = element.claim;
         return true;
     }
     return false;
 }
 
+// name can be setted explicitly
```
tiger5226: typo

src/rpc/claimtrie.cpp
```diff
@@ -102,91 +107,194 @@ std::string escapeNonUtf8(const std::string& name)
...
+            result.pushKV(T_VALUE, value);
+
+        CTxDestination address;
+        if (ExtractDestination(coin.out.scriptPubKey, address))
+            result.pushKV(T_ADDRESS, EncodeDestination(address));
+    }
+
+    result.pushKV(T_CLAIMID, claim.claimId.GetHex());
+    result.pushKV(T_TXID, claim.outPoint.hash.GetHex());
+    result.pushKV(T_N, (int)claim.outPoint.n);
```
tiger5226: should we cast this?

src/rpc/claimtrie.cpp
```diff
@@ -102,91 +107,194 @@ std::string escapeNonUtf8(const std::string& name)
...
+        if (extractValue(coin.out.scriptPubKey, value))
+            ret.pushKV(T_VALUE, value);
+
+        CTxDestination address;
+        if (ExtractDestination(coin.out.scriptPubKey, address))
+            ret.pushKV(T_ADDRESS, EncodeDestination(address));
+    }
+
+    ret.pushKV(T_TXID, support.outPoint.hash.GetHex());
+    ret.pushKV(T_N, (int)support.outPoint.n);
```
tiger5226: casting again, maybe the type needs to be updated?

src/rpc/claimtrie.cpp
```diff
@@ -102,91 +107,194 @@ std::string escapeNonUtf8(const std::string& name)
...
+    result.pushKV(T_SUPPORTS, supportObjs);
+
+    return result;
+}
+
 bool validParams(const UniValue& params, uint8_t required, uint8_t optional)
 {
     auto count = params.size();
-    return count == required || count == required + optional;
+    return count >= required && count <= required + optional;
```
tiger5226: nice catch!

bvbfan:
Changes are breaking since i unify outputs which makes changes to the fields.

src/rpc/claimrpchelp.h
bvbfan: I think is more readable now :) You see it on one place and there is no need to verify it's same or not and it's correct on every place or not.

src/rpc/claimtrie.cpp
bvbfan: UniValue does not provide an uint32 constructor which can lead to ambiguous.

src/rpc/claimrpchelp.h
tiger5226: I was also looking at it through the PR lens. Will look at it in the code base too.

bvbfan: Overall code base is way better to me, logic and help message are separate which makes logic to be easy follow, OTOH help message are minimize.

BrannonKing:
The cli help (that shows all the available methods) isn't showing the parameters for the methods in rpc/claimtrie.cpp. It's like the argNames are ignored in that file.

PR #312, opened 2019-09-04
initial commit of metadata on supports

### LINKED ISSUE #272: add optional metadata for supports
This is prototyped in the metadata_for_supports branch. This sample implementation doesn't require a hard-fork in the code, but it probably needs one to ensure that enough people get the change. If we could instead get the metadata into the witness section of the block, we may be able to make this usable without a hard fork. See #273 

This is to enable options for advertisements and certifications and content reviews.

BrannonKing:
The nonstandard transactions themselves should be okay in the current code; we just don't want a different hash or trie computation until the fork height.

Fixes #272

### DIFF
--- a/src/nameclaim.cpp
+++ b/src/nameclaim.cpp
@@ -37,11 +37,17 @@ CScript ClaimNameScript(std::string name, std::string value, bool fakeSuffix)
     return ret;
 }
 
-CScript SupportClaimScript(std::string name, uint160 claimId, bool fakeSuffix)
+CScript SupportClaimScript(std::string name, uint160 claimId, std::string value, bool fakeSuffix)
 {
     std::vector<unsigned char> vchName(name.begin(), name.end());
     std::vector<unsigned char> vchClaimId(claimId.begin(), claimId.end());
-    auto ret = CScript() << OP_SUPPORT_CLAIM << vchName << vchClaimId << OP_2DROP << OP_DROP;
+    CScript ret;
+    if (value.empty())
+        ret = CScript() << OP_SUPPORT_CLAIM << vchName << vchClaimId << OP_2DROP << OP_DROP;
+    else {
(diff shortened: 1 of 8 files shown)


src/nameclaim.cpp
```diff
@@ -99,35 +106,42 @@ bool DecodeClaimScript(const CScript& scriptIn, int& op, std::vector<std::vector
...
         }
     }
-    if (op == OP_UPDATE_CLAIM)
+
+    if (!scriptIn.GetOp(pc, opcode, vchParam3))
+    {
+        return false;
+    }
+    auto last_drop = OP_DROP;
+    if (opcode >= 0 && opcode <= OP_PUSHDATA4 && op != OP_CLAIM_NAME)
```
bvbfan: It does not look correct or we miss OP_SUPPORT_CLAIM && !allowSupportMetadata it's same as OP_CLAIM_NAME ?

bvbfan:
```
switch (op) {
    case OP_SUPPORT_CLAIM:
        if (allowSupportMetadata)
        // fallthrough
    case OP_UPDATE_CLAIM: {
        if (opcode < 0 || opcode > OP_PUSHDATA4)
            return false;
        if (!scriptIn.GetOp(pc, opcode))
            return false;
    } // fallthrough
    case OP_CLAIM_NAME:
        if (opcode != OP_2DROP)
           return false;
    break;
}
```

BrannonKing: I don't understand your concern here. The logic is "if not a claim op and have another data chunk", that's fine -- that's what we want. That third data is required on the update op but not required on the support op.

bvbfan: I was afraid on op == OP_SUPPORT_CLAIM and allowSupportMetadata == false we can enter in by the condition but opcode is OP_2DROP so it's OK.

PR #318, opened 2019-09-17
Fix unit tests after recent breaks.

lbrynaut:


BrannonKing:
I can understand maintaining the reqSigs order. What is the purpose of the noop test?

BrannonKing:
I understand now; it's for the `make check`.

PR #319, opened 2019-09-19
Show claimtrie help in cli rpc

bvbfan:


BrannonKing:
Bid and sequence are optional parameters, no?

PR #321, opened 2019-09-27
reverted base-trie-on-disk

BrannonKing:
It wasn't substantially better so I yanked it. It was theoretically better, but it didn't look better by the numbers. It's going to take a substantial number of claims to make the capped leveldb approach better, and before that happens we'll have done the sqlite backing.

### DIFF
--- a/src/claimtrie.cpp
+++ b/src/claimtrie.cpp
@@ -271,8 +271,8 @@ bool CClaimTrieCacheBase::haveClaim(const std::string& name, const COutPoint& ou
         return true;
     if (it || nodesToDelete.count(name))
         return false;
-    CClaimTrieData data;
-    return base->find(name, data) && data.haveClaim(outPoint);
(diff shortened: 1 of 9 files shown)


BrannonKing:
For review, I want to note the previous copy method:```
void CClaimTrieCacheHashFork::copyAllBaseToCache()
{
    recurseNodes({}, [this](const std::string& name, const CClaimTrieData& data) {
        if (nodesAlreadyCached.insert(name).second)
            nodesToAddOrUpdate.insert(name, data);
    });

    for (auto it = nodesToAddOrUpdate.begin(); it != nodesToAddOrUpdate.end(); ++it) {
        it->hash.SetNull();
        it->flags |= CClaimTrieDataFlags::HASH_DIRTY;
    }
}```

BrannonKing:
removing in favor of memory branch

PR #344, opened 2019-10-22
Implement basic log rotation (closes #211)

### LINKED ISSUE #211: lbrycrdd should have a logroll if its printing console output to logs.
Since it prints every line about every bit of indexing that it does it should be able to roll the log file a bit so that you don't end up with a full blockchain sized log file.

gahag:
As discussed in #211, this PR introduces the automatic moving of and existing log file to an `.old` file. This means logs are preserved for the last 2 executions, and also implies the -shrinkdebugfile is no longer useful, and thus removed.

Please tell me if there are any improvements to be made.


BrannonKing:
How does the rename handle conflicts should an existing "old" file already exist?

gahag:
The old file is always replaced if it exists. This is an important part for assuring you have the log from the last two executions, but no more.

gahag:
@BrannonKing please tell me if there's anything to be improved on this.

PR #364, opened 2020-01-28
Use fully static linkage

bvbfan:


bvbfan:
According to https://pkgs.org/download/libicu icu is distributed without static libs (some distros have a static libs package) which means at least custom compilation of icu is needed for correct linkage.

BrannonKing:
Not going down this road at the moment

PR #366, opened 2020-02-06
Fix blockdir beamer

tiger5226:


.travis.yml
```diff
@@ -24,11 +24,13 @@ jobs:
         - mkdir -p ${HOME}/ccache
         - docker pull $DOCKER_BUILD_IMAGE
       script:
+        - echo "build..."
         - docker run -v "$(pwd):/lbrycrd" -v "${HOME}/ccache:/ccache" -w /lbrycrd -e CCACHE_DIR=/ccache ${DOCKER_IMAGE} packaging/build_${NAME}_64bit.sh
       before_deploy:
         - mkdir -p dist
         - sudo zip -Xj dist/lbrycrd-${NAME}.zip src/lbrycrdd${EXT} src/lbrycrd-cli${EXT} src/lbrycrd-tx${EXT}
         - sudo zip -Xj dist/lbrycrd-${NAME}-test.zip src/test/test_lbrycrd${EXT} src/test/test_lbrycrd_fuzzy${EXT}
+        - sudo cp dist/lbrycrd-${NAME}.zip packaging/docker-for-binary/lbrycrd-${NAME}.zip
```
BrannonKing: should be `ln -s` ?

.travis.yml
```diff
@@ -44,6 +46,13 @@ jobs:
           on:
             repo: lbryio/lbrycrd
             all_branches: true
+        - provider: script
+          script: bash packaging/docker-for-binary/auto_deploy.sh ${TRAVIS_BRANCH}
+          skip_cleanup: true
+          on:
+            repo: lbryio/lbrycrd
+            all_branches: true
```
BrannonKing: This has to be tags only. (You're welcome to tag things in the lbrycrd repo as you see fit.)

.travis.yml
tiger5226: I intended to copy the file so docker grabs it for the build context. I did this instead of moving it so I did not interfere with any other processes or tests that might do something with it. I am not the best with unix commands, if you think a symbolic link is better I can update and push a change. I did not think to do that. I did not know that would work with docker either.

PR #375, opened 2020-03-30
Add hard fork to include more claim data

bvbfan:
It's future hard fork request to ensure:

1. Include claim name, bid and sequence to the merkle hash
2. Allow only UTF-8 like claim names
3. Disallow following characters `=&#:$@%?;\/"<>*\n\t\r\b\0` in claim name

Adjust consensus parameter `nClaimInfoInMerkleForkHeight` to allow it in production as well as `nMaxRemovalWorkaroundHeight` to purge out implementation detail bug which allow addition of claim without delay, see comments for more details.


BrannonKing:
The original list of that must be included in the fork:
1. Fork out the removal workaround (which we tried to fork out previously and failed).
2. Make names with nulls in them not be accepted.
    1. And what about these symbols? *** ? _ % : # $ \n \t \r \b & = /**
    2. Enforce that any new claim names are valid UTF-8.
3. IMPORTANT: add the name, bid, and sequence numbers into the claim value hash. 
    1. The user should not have to trust that the parameters to the proof RPC methods were used correctly.
4. Increase the maximum metadata size from 8KB to ?

It was decided that item 2 could be done by enforcing it at the mempool level only. The name enforcement does not need to become part of consensus. Hence, it technically doesn't need a fork.

It appears that item 1 is not in this changeset? Actually, it looks like we simply need to set the nMaxRemovalWorkaroundHeight. (We also need to set the other effective heights for this fork.)

The maximum metadata size should probably be doubled. No more than that.

PR #376, opened 2020-04-01
adjust testnet chain params to prepare for reset

nikooo777:


src/chainparams.cpp
```diff
@@ -247,24 +247,26 @@ class CTestNetParams : public CChainParams {
...
-        consensus.nAllowMinDiffMinHeight = 277299;
-        consensus.nAllowMinDiffMaxHeight = 1100000;
-        consensus.nNormalizedNameForkHeight = 993380;   // targeting, 21 Feb 2019
+        consensus.nExtendedClaimExpirationForkHeight = 1;
+        consensus.nAllowMinDiffMinHeight = 1;
+        consensus.nAllowMinDiffMaxHeight = 2;
+        consensus.nNormalizedNameForkHeight = 1;   // targeting, 21 Feb 2019
         consensus.nMinRemovalWorkaroundHeight = 99;
-        consensus.nMaxRemovalWorkaroundHeight = 100000000;
-        consensus.nWitnessForkHeight = 1198600;
```
BrannonKing, line 266: We don't want to delete this one!. It needs to be set, probably to block 120.

src/chainparams.cpp
```diff
@@ -273,21 +275,11 @@ class CTestNetParams : public CChainParams {
         consensus.vDeployments[Consensus::DEPLOYMENT_TESTDUMMY].nStartTime = 1199145601; // January 1, 2008
         consensus.vDeployments[Consensus::DEPLOYMENT_TESTDUMMY].nTimeout = 1230767999; // December 31, 2008
 
-        // Deployment of BIP68, BIP112, and BIP113.
```
BrannonKing, line 276: There's no need to delete these DEPLOYMENT bits.

src/chainparams.cpp
nikooo777: wooops! didn't meant to get rid of it. fixed

src/chainparams.cpp
nikooo777: aaand fixed as well

PR #397, opened 2021-01-10
Fix placeholders for boost bind (>=1.73.0)

Quasido:
Needed to compile against system libraries on Manjaro.
I've tested that it still compiles with boost 1.72.0 (prev Manjaro package), however I have not tested with the docker image.


BrannonKing:
Have you looked at the v17_master and v19_master branches? Do they have this same issue?

Quasido:
Sorry didn't check for other branches. Fixed in v19.

PR #405, opened 2021-07-14
add endpoint that returns a block's hash and height

jeffreypicard:
In the sdk we just use the height so I added this for convenience

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -447,11 +447,13 @@ void SetupServerArgs()
 
 #if ENABLE_ZMQ
     gArgs.AddArg("-zmqpubhashblock=<address>", "Enable publish hash block in <address>", false, OptionsCategory::ZMQ);
+    gArgs.AddArg("-zmqpubhashblockheader=<address>", "Enable publish hash block in <address>", false, OptionsCategory::ZMQ);
     gArgs.AddArg("-zmqpubhashtx=<address>", "Enable publish hash transaction in <address>", false, OptionsCategory::ZMQ);
     gArgs.AddArg("-zmqpubrawblock=<address>", "Enable publish raw block in <address>", false, OptionsCategory::ZMQ);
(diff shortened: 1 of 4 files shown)


src/zmq/zmqpublishnotifier.cpp
```diff
@@ -156,6 +157,18 @@ bool CZMQPublishHashBlockNotifier::NotifyBlock(const CBlockIndex *pindex)
...
 
+bool CZMQPublishHashBlockHeaderNotifier::NotifyBlock(const CBlockIndex *pindex)
+{
+    uint256 hash = pindex->GetBlockHash();
+    uint32_t height = pindex->nHeight;
+    LogPrint(BCLog::ZMQ, "zmq: Publish hashblock %s\n", hash.GetHex());
+    char data[36];
+    for (unsigned int i = 0; i < 32; i++)
+        data[31 - i] = hash.begin()[i];
+    memcpy(data + 32, (char *)&height, 4);
```
BrannonKing, line 168: Copying an integer this way is architecture-dependent. Look at the functions in the endian.h file.

src/init.cpp
```diff
@@ -447,11 +447,13 @@ void SetupServerArgs()
 
 #if ENABLE_ZMQ
     gArgs.AddArg("-zmqpubhashblock=<address>", "Enable publish hash block in <address>", false, OptionsCategory::ZMQ);
+    gArgs.AddArg("-zmqpubhashblockheader=<address>", "Enable publish hash block in <address>", false, OptionsCategory::ZMQ);
```
BrannonKing, line 450: The comment wasn't changed. And the name implies that you are sending the entire block header -- more than just hash and height.

BrannonKing (changes requested):
The entire header is 112 bytes. I think it would be useful to have a method that publishes that in full. The CBlockIndex has a method to get a CBlockHeader; that header is serializable, if that helps.

Issue #363, opened 2020-01-27
percentage markers for loading the block index and validating the claimtrie


BrannonKing:
Those two things tend to stall a bit on startup. It would be nice to have the [0%]... output for them.

PR #408, opened 2021-09-15
Add progress logging to claimtrie loading and validation.

pseudoscalar:
To resolve [363](https://github.com/lbryio/lbrycrd/issues/363)

### DIFF
--- a/src/claimtrie.cpp
+++ b/src/claimtrie.cpp
@@ -442,8 +442,17 @@ uint256 recursiveMerkleHash(TIterator& it, const iCbType<TIterator>& process)
 bool CClaimTrieCacheBase::recursiveCheckConsistency(CClaimTrie::const_iterator& it, std::string& failed) const
 {
     struct CRecursiveBreak {};
+    int nodesVisited = 0;
+    int reportDone = 0;
     using iterator = CClaimTrie::const_iterator;
-    iCbType<iterator> process = [&failed, &process](iterator& it) {
+    iCbType<iterator> process = [this, &reportDone, &nodesVisited, &failed, &process](iterator& it) {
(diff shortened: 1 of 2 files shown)


src/claimtrie.cpp
```diff
@@ -576,11 +586,28 @@ bool CClaimTrieCacheBase::ReadFromDisk(const CBlockIndex* tip)
     base->clear();
     boost::scoped_ptr<CDBIterator> pcursor(base->db->NewIterator());
 
+    LogPrintf("[0%%]..."); /* Continued */
+    int totalRecords = 0;
     for (pcursor->SeekToFirst(); pcursor->Valid(); pcursor->Next()) {
+        totalRecords++;
+    }
```
BrannonKing, line 593: This pre-count loop could be very expensive, as the disk cache won't be warm. How long is this loop taking compared to the lower loop?

pseudoscalar: I was seeing 1-3 seconds in the precount loop vs ~2 minutes for the lower loop.

pseudoscalar: Sorry, I should clarify. ~2 minutes for the whole load from disk. It was split pretty evenly between the 2 lower loops (so ~1 minute each). That's why I decided to do a precount loop: counting the total in the first lower loop would leave the operation already half done (time-wise). I couldn't see another way to estimate progress, and the added time seemed negligible next to the main operation.

pseudoscalar: Did you want me to find a way to eliminate the precount, or is this good?

BrannonKing:
Have you looked at adding the progress notifiers to the v17_master and v19_master branches?

pseudoscalar:
I haven't yet. I start putting together some PRs for those.

